PR #2, opened 2021-03-18
DM-27485: Design cp_verify

czwa:
This is the initial design ticket for what I hope will be a sufficiently modular package to verify calibration products.  The main concept is to calculate some metric from amplifiers, full detectors, and/or catalog products, check those metrics against tests defined in DMTN-101 (although that is not complete either), and then merge those results for an entire set of test exposures.  If the calibration is good, all the tests should pass, and it can be certified.  If some tests fail, they should provide some guidance on what needs to be corrected.


python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+__all__ = ['CpVerifyExpMergeConfig', 'CpVerifyExpMergeTask',
+           'CpVerifyRunMergeConfig', 'CpVerifyRunMergeTask']
+
+
+class CpVerifyExpMergeConnections(pipeBase.PipelineTaskConnections,
+                                  dimensions={"instrument", "exposure"},
+                                  defaultTemplates={}):
+    inputStats = cT.Input(
```
taranu, line 33: I'll note here and not repeat myself: the style guide recommends snake_case for public member names in primarily new code, which I think this package qualifies as.

README.md
```diff
@@ -0,0 +1,4 @@
+Description
+===========
+
+The ip_isr package provides Instrument Signature Removal related tasks. ISR includes steps such as combining multiple amplifiers into one full CCD image,  corrections for overscans, crosstalk, bias and dark frames, and the creation of variance and mask planes.
```
laurenam: Need to update the copy-pasta!

(...and I'd rather like this before I get too far into this review so that I know what the package is advertising!)

python/lsst/__init__.py
```diff
@@ -0,0 +1,2 @@
+import pkgutil, lsstimport
```
laurenam: Is the `lsstimport` necessary here? E.g.,
https://github.com/lsst/ctrl_bps/commit/7a053e5d079d4d35af75ba5fa1d754f2bc6af673#diff-ac9066211ffc771ae1538066f8f10df97432e0c6fdcd55925e408b35c7ff9ee2
and
https://github.com/lsst/daf_butler/commit/450124b8c5bd62f7f7954a08e63dafbe09ea8268

python/lsst/cp/verify/__init__.py
```diff
@@ -0,0 +1,10 @@
+from .version import *
+from .verifyStats import *
+from .mergeResults import *
+
+from .verifyBias import *
+from .verifyDark import *
+from .verifyDefects import *
+
+import pkgutil, lsstimport
+__path__ = pkgutil.extend_path(__path__, __name__)
```
laurenam: Is this `import` & `__path__` setting needed here?  I don't see it in other packages...

python/lsst/cp/verify/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+def mergeStatDict(statDictA, statDictB):
+    """Merge one dictionary of statistics with another.
+
+    Parameters
+    ----------
+    statDictA : `dict` [`str`, `dict`]
+        A dictionary of statistics.
+    statDictB : `dict` [`str`, `dict`]
+        Another dictionary of statistics.
```
laurenam: Some comment about the fact that if an `amp` keys in common get the `statDictB` values?  (is this dangerous?)

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+class CpVerifyBiasConfig(CpVerifyStatsConfig,
+                         pipelineConnections=CpVerifyStatsConnections):
+    """Inherits from base CpVerifyStatsConfig
+    """
+
+    def setDefaults(self):
+        super().setDefaults()
+        self.imageStatKeywords = {'MEAN': 'MEAN',  # noqa F841
+                                  'NOISE': 'STDEVCLIP', }
```
laurenam, line 37: Would fit on a single line within 110 line-width here...or are you really trying to facilitate/anticipate additions?

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.crImageStatKeywords = {'CR_NOISE': 'STDEV', }  # noqa F841
+
+
+class CpVerifyBiasTask(CpVerifyStatsTask):
+    """
+    """
```
laurenam, line 44: Summarize me shortly, please!

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+from .verifyStats import CpVerifyStatsConfig, CpVerifyStatsTask, CpVerifyStatsConnections
+
+
+__all__ = ['CpVerifyBiasConfig', 'CpVerifyBiasTask']
+
+
+class CpVerifyBiasConfig(CpVerifyStatsConfig,
+                         pipelineConnections=CpVerifyStatsConnections):
+    """Inherits from base CpVerifyStatsConfig
```
laurenam: Period.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    """
+    ConfigClass = CpVerifyBiasConfig
+    _DefaultName = 'cpVerifyBias'
+
+    def verify(self, exposure, statisticsDictionary):
+        """Verify if the measured statistics meet the verification criteria.
+
+        Parameters
+        ----------
```
laurenam, line 52: Document `exposure`.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class CpVerifyBiasTask(CpVerifyStatsTask):
+    """
+    """
+    ConfigClass = CpVerifyBiasConfig
+    _DefaultName = 'cpVerifyBias'
+
+    def verify(self, exposure, statisticsDictionary):
+        """Verify if the measured statistics meet the verification criteria.
```
laurenam: I don't think you can verify "if" something..."whether/that".

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Parameters
+        ----------
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
+            Dictionary of measured statistics.
```
laurenam: It would help if you state what the various keys & sub-keys represent.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, `bool`]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the verification criteria.
+        success : `bool`
+            A boolean indicating if all tests have passed.
```
laurenam, line 67: "if" -> "whether [or not]"?

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+
```
laurenam: No newline.
https://developer.lsst.io/python/numpydoc.html#docstrings-of-methods-and-functions-should-not-be-preceded-or-followed-by-a-blank-line

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+        """Verify if the measured statistics meet the verification criteria.
+
+        Parameters
+        ----------
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
+            Dictionary of measured statistics.
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, `bool`]]
```
laurenam: Might be best to be consistent in your use of "stats" vs. "statistics" vs. "stat" in variable names.  Also "dict" vs. "dictionary".  (I know there is a strong team that is against abbreviations -- one I often play for, but in these two cases, I'm a-ok with the abbreviated forms...so long as you are consistent or else I'll start trying to figure out what distinction is being made!)

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+        detector = exposure.getDetector()
+        ampStats = statisticsDictionary['AMP']
+
+        verifyStats = dict()
+        success = True
+        for ampName, stats in ampStats.items():
+            verify = dict()
+
+            # Test 4.2: Mean is 0.0:
```
laurenam: What do these test #s refer to?

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+            verify['MEAN'] = (np.abs(stats['MEAN']) < stats['NOISE'])
```
laurenam: I don't see how this ensures "Mean is 0.0"? (add "to within statistical error")?

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+class CpVerifyExpMergeConfig(pipeBase.PipelineTaskConfig,
+                             pipelineConnections=CpVerifyExpMergeConnections):
+    """Configuration paramters for exposure stats merging.
+    """
+    exposureStatKeywords = pexConfig.DictField(
+        keytype=str,
+        itemtype=str,
+        doc="Statistics to run on the set of exposure values.",
```
taranu: What are the keys and values? It doesn't seem obvious to me. I presume that the statistic is the value.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+            calcStats = dict()
+
+            if detStats['SUCCESS'] is True:
+                calcStats['SUCCESS'] = True
+            else:
+                calcStats['SUCCESS'] = False
+                calcStats['FAILURES'] = list()
+                success = False
+                # See if the detector failed
+                if 'DET' in detStats['VERIFY']:
```
taranu, line 143: You could write `for testname in detStats.get('VERIFY', ()):`, though I'm not sure if that's as readable.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+                    for testName in detStats['VERIFY']['DET']:
+                        if detStats['VERIFY']['DET'][testName] is False:
+                            calcStats['FAILURES'].append(testName)
+                # See if the catalog failed
+                if 'CATALOG' in detStats['VERIFY']:
+                    for testName in detStats['VERIFY']['CATALOG']:
+                        if detStats['VERIFY']['CATALOG'][testName] is False:
+                            calcStats['FAILURES'].append(testName)
+                # See if an amplifier failed
+                for ampName in detStats['VERIFY']['AMP']:
```
taranu: Is this not a dict or is there another reason you're not writing `for ampName, ampValue in detStats['VERIFY']['AMP'].items()`?

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+        return pipeBase.Struct(
+            outputStats=outputStats,
+        )
+
+    def verify(self, statisticsDictionary):
+        """Verify if the measured statistics meet the verification criteria.
+
+        Parameters
+        ----------
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
```
taranu, line 177: Is scalar a thing in numpydoc? Are you using it as shorthand for "`float` or `int`" or something else?

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+class CpVerifyRunMergeConfig(pipeBase.PipelineTaskConfig,
+                             pipelineConnections=CpVerifyRunMergeConnections):
+    """Configuration paramters for exposure stats merging.
+    """
+    runStatKeywords = pexConfig.DictField(
+        keytype=str,
+        itemtype=str,
+        doc="Run statistics to create.",
```
taranu: This doc string is even more sparse than the one for `exposureStatKeywords` and could use more elaboration.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+          ExposureId2:
+            [...]
+          MEAN_UNIMODAL: boolean
+          SIGMA_UNIMODAL: boolean
+        """
+        outputStats = dict()
+        success = True
+        for expStats, dimensions in zip(inputStats, inputDims):
+            expId = dimensions['exposure']
+            calcStats = dict()
```
taranu: The internet occasionally nags me to initialize tuple/list/dict from literals instead of constructors, so I'm here to pass it on.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+        return pipeBase.Struct(
+            outputStats=outputStats,
+        )
+
+    def verify(self, statisticsDictionary):
+        """Verify if the measured statistics meet the verification criteria.
+
+        Parameters
+        ----------
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
```
taranu: See the note on scalar above.

python/lsst/cp/verify/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+__all__ = ['mergeStatDict']
+
+
+def mergeStatDict(statDictA, statDictB):
+    """Merge one dictionary of statistics with another.
+
+    Parameters
+    ----------
+    statDictA : `dict` [`str`, `dict`]
```
taranu: You can merge these two if you like: https://developer.lsst.io/python/numpydoc.html#shorthand

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+
+            # Test 4.4: CR rejection matches clipped mean
+            verify['CR_NOISE'] = (np.abs(stats['NOISE'] - stats['CR_NOISE'])/stats['CR_NOISE'] <= 0.05)
+
+            verify['SUCCESS'] = bool(np.all(list(verify.values())))
+            if verify['SUCCESS'] is False:
+                success = False
+
+            # Why is this necessary?  Why are these booleans not
+            # boolean enough for yaml.safe_dump?
```
taranu: The MEAN, NOISE and CR_NOISE ones are np.bool, not bool. You should cast them like for SUCCESS.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,95 @@
...
+    def setDefaults(self):
+        super().setDefaults()
+        self.imageStatKeywords = {'MEAN': 'MEAN',  # noqa F841
+                                  'NOISE': 'STDEVCLIP', }
+
+        self.crImageStatKeywords = {'CR_NOISE': 'STDEV', }  # noqa F841
+
+
+class CpVerifyBiasTask(CpVerifyStatsTask):
+    """
```
taranu: It's probably worth putting a five-word docstring even if it tells you nothing more than the name of the class does.

python/lsst/cp/verify/verifyDark.py
```diff
@@ -0,0 +1,93 @@
...
+            # Test 5.4: CR rejection matches clipped mean
+            verify['CR_NOISE'] = (np.abs(stats['NOISE'] - stats['CR_NOISE'])/stats['CR_NOISE'] <= 0.05)
+
+            verify['SUCCESS'] = bool(np.all(list(verify.values())))
+            if verify['SUCCESS'] is False:
+                success = False
+
+            # Why is this necessary?  Why are these booleans not
+            # boolean enough for yaml.safe_dump?
+            verify = {key: bool(value) for key, value in verify.items()}
```
taranu: See above.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -0,0 +1,142 @@
...
+            ampExp = exposure.Factory(exposure, amp.getBBox())
+
+            normImage = ampExp.getImage()
+            normArray = normImage.getArray()
+
+            normArray -= outputStats[ampName]['MEDIAN']
+            normArray /= outputStats[ampName]['STDEV']
+
+            probability = scipy.stats.norm.pdf(normArray)
+            trials = np.prod(probability.shape)
```
taranu: `probability.size`

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -0,0 +1,142 @@
...
+
+            verify['PROB_TEST'] = True if stats['STAT_OUTLIERS'] == stats['DEFECT_PIXELS'] else False
+
+            verify['SUCCESS'] = bool(np.all(list(verify.values())))
+            if verify['SUCCESS'] is False:
+                success = False
+
+            # Why is this necessary?  Why are these booleans not
+            # boolean enough for yaml.safe_dump?
+            verify = {key: bool(value) for key, value in verify.items()}
```
taranu: It doesn't look like this should be necessary here since they're all ternary bools above.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -0,0 +1,142 @@
...
+        success : `bool`
+            A boolean indicating if all tests have passed.
+        """
+        ampStats = statisticsDictionary['AMP']
+        verifyStats = dict()
+        success = True
+        for ampName, stats in ampStats.items():
+            verify = dict()
+
+            # These are not defined yet.
```
taranu: What are not defined? I'm not sure what you're referring to.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -0,0 +1,142 @@
...
+            if stats['UNMASKED_OUTLIERS'] > stats['OUTLIERS']:
```
taranu: Why isn't this a ternary statement like the others?

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -0,0 +1,142 @@
...
+                verify['OUTLIERS'] = True
+            else:
+                verify['OUTLIERS'] = False
+
+            verify['MIN'] = True if stats['MIN'] >= stats['UNMASKED_MIN'] else False
```
taranu: Are you trying to guard against NaNs (and if so, why aren't you doing so in the other verify methods), or is this just in place of a bool cast?

tests/test_mergeResults.py
```diff
@@ -0,0 +1,144 @@
...
+        """Generate simulated verify statistics, and prove they merge
+        correctly.
+        """
+
+        expMergeTask = cpVerify.CpVerifyExpMergeTask()
+
+        expResults1 = expMergeTask.run(self.detectorResults1, self.camera, self.detectorDimensions1)
+        expResults2 = expMergeTask.run(self.detectorResults2, self.camera, self.detectorDimensions2)
+        self.assertTrue(expResults1.outputStats['SUCCESS'])
+        self.assertFalse(expResults2.outputStats['SUCCESS'])  # Expecte to have one failure
```
taranu: Expected?

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,323 @@
...
+                detector id value (`int`)
+
+        Returns
+        -------
+        outputStats : `dict`
+            Merged full exposure statistics.
+
+        Notes
+        -----
+        The outputStats should have a yaml representation of the form:
```
taranu, line 114: Is it not possible to make this a dataclass with a `write_yaml` method or something? It would add a little more boilerplate but is much more easily documented than multiply-nested dicts.

taranu (commented):
Most comments are pedantic but a few might be useful.

python/lsst/cp/verify/mergeResults.py
czwa: Clarified.  The key is the name to be recorded in the output verify yaml, and the value is the afwMath statistic name.

python/lsst/cp/verify/mergeResults.py
czwa: No.  That makes it more readable, so I've made that change.  I've also updated in the interior for loops to do similar things.

python/lsst/cp/verify/mergeResults.py
czwa: It's shorthand for "there is no more recursive of dicts or lists".  I've removed the inner dict type information, and moved that to the docstring as suggested by the docstring guide.

python/lsst/cp/verify/mergeResults.py
czwa: I have changed all the ones I can remember in this package.

python/lsst/cp/verify/verifyDefects.py
czwa: Not defined in DMTN-101.  This package (as it will say in the README.md once I push change) will implement all of the tests defined there, but the defect section is currently unwritten.  I wanted this first pass at the package to be reviewed and merged before sitting down to define everything that has been left undone.  I've updated the comment to point to the DMTN-101 definition.

python/lsst/cp/verify/verifyDefects.py
czwa: I'm 99% sure it's because I wrote that one day, and the rest a different day, and never unified styles.

python/lsst/cp/verify/verifyDefects.py
czwa: Casting to bool is more reasonable, so I've switched to that.  The problem of not fully designing a test before trying to implement it.

python/lsst/cp/verify/mergeResults.py
czwa: The short answer is yes, the long answer is that I've added a ticket to do that: https://jira.lsstcorp.org/browse/DM-29643

python/lsst/cp/verify/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Returns
+    -------
+    output : `dict` [`str`, `dict`]
+        A dictionary containing the union of the internal
+        dictionaries of ``statDictA`` and ``statDictB``, indexed
+        by the amplifier names.
+    """
+    for amp in statDictA.keys():
+        statDictA[amp].update(statDictB[amp])
```
laurenam, line 51: So this assumes that the list of `amp` keys is the same for both `statDict`s?  Does that warrant a comment (& possibly a check)?

.gitignore
```diff
@@ -0,0 +1,23 @@
...
+*.cfgc
+*.pyc
+*.xml
+doc/*.inc
+doc/*.tag
+doc/doxygen.conf
+doc/html
+doc/xml
+doc/_build
+doc/py-api/
```
laurenam, line 17: I suppose it doesn't hurt, but you don't actually have a`doc/` directory.

README.md
```diff
@@ -0,0 +1,6 @@
+Description
+===========
+
+The cp_verify package provides tasks to test the quality of calibration products.  These tasks can be run after calibration production to perform the tests defined in DMTN-101, and collate the results by exposure and for the full calibration run.  Calibrations that successfully pass all tests can then be certified without needing an extensive review.
+
+In addition, these tasks can be used to generate metrics to monitor the stability of the camera over time by applying the same tests to newly taken raw calibration frames.
```
laurenam, line 6: I might be helpful to add a list of the tasks provided here and perhaps include some simple examples of how to run them (but perhaps this is planned for a future `doc/` section addition)?

setup.cfg
```diff
@@ -0,0 +1,10 @@
+[flake8]
+max-line-length = 110
+ignore = E133, E226, E228, N802, N803, N806, E266, N812, N815, N816, W503
+# TODO: remove E266 when Task documentation is converted to rst in DM-14207.
```
laurenam: Seems unnecessary for this new package?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        name="detectorStats",
+        doc="Output statistics from cp_verify.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "exposure", "detector"],
+    )
+
+
+class CpVerifyStatsConfig(pipeBase.PipelineTaskConfig,
+                          pipelineConnections=CpVerifyStatsConnections):
+    """Configuration parameters for CpVerifyTask
```
laurenam: CpVerify**Stats**Task?  And period.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        dtype=bool,
+        doc="Normalize by exposure time?",
+        default=False,
+    )
+
+    doCR = pexConfig.Field(
+        dtype=bool,
+        doc="Run CR rejection?",
+        default=False,
+    )
```
laurenam, line 87: If this is meant to test generated "master" calibration frames, shouldn't any CR rejection that can be done already have been (and allowing for doing it again here would thus result in these tests operating on --erroneously-- modified versions of the calibs)?  Oh, but a PSF is required for the CR finding code, so you are setting a dummy PSF (defined by the config pars below). I guess I'm missing the point of validating modified calibs as opposed to "as constructed".  If this dummy-based CR rejection is reasonable, why not do it in the calib construction in the first place? [Apologies...I'm clearly missing something here, likely just due to my naïveté when it comes to calibration construction procedures!]

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+    psfFwhm = pexConfig.Field(
+        dtype=float,
+        default=3.0,
+        doc="Repair PSF FWHM (pixels).",
+    )
+    psfSize = pexConfig.Field(
+        dtype=int,
+        default=21,
+        doc="Repair PSF size (pixels).",
```
laurenam: Having trouble with `psfFwhm` vs. `psfSize`.  I believe the latter is the size of the PSF kernel `bbox`, but the name does not make that clear...

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+    useReadNoise = pexConfig.Field(
+        dtype=bool,
+        doc="Compare sigma against read noise?",
+        default=True,
+    )
+
+    numSigmaClip = pexConfig.Field(
+        dtype=float,
+        doc="Rejection threshold (sigma) for statistics clipping.",
+        default=3.0,
```
laurenam: https://dmtn-101.lsst.io/DMTN-101.pdf quotes 5-sigma throughout...

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+              TEST2: boolean
+            Amp2:
+            Amp3:
+        SUCCESS: boolean
+
+        """
+        outputStats = {}
+
+        if self.config.doVignette:
+            VignetteExposure(inputExp, doUpdateMask=True, maskPlane='BAD',
```
laurenam: If you really do mean to modify the calib you're validating, the mask plane for the vignetted region should be `NO_DATA` (and I'm working on unifying that on DM-29326 which may moot DM-29325!)

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+            raise RuntimeError("No detector found in exposure!")
+
+        for amp in detector.getAmplifiers():
+            outputStats[amp.getName()] = {}
+
+        return outputStats
+
+    # Image measurement methods.
+    def imageStatistics(self, exposure, statControl):
+        """Measure all types of image statistics.
```
laurenam: Really **all** of the statistics?!

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+            Statistics control object with the parameters defined by
+            the config.
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, scalar]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the statistics measured and their values.
+        """
+
```
laurenam, line 303: Blank line not needed.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
+            Dictionary of measured statistics.
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, `bool`]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the verification criteria.
+        success : `bool`
+            A boolean indicating if all tests have passed.
```
laurenam: whether (or not)

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+    # Methods that need to be implemented by the calibration-level subclasses.
+    def catalogStatistics(self, exposure, statControl):
+        """Calculate statistics from a catalog.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure to measure.
+        statControl : `lsst.afw.math.StatControl`
+            Statistics control object with the parameters defined by
```
laurenam: Don't need "the".

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        dtype=bool,
+        doc="Mask vignetted regions?",
+        default=False,
+    )
+    doNormalize = pexConfig.Field(
+        dtype=bool,
+        doc="Normalize by exposure time?",
+        default=False,
+    )
+
```
laurenam, line 81: If space here is to delineate different flavors of config parameters, might as well make the delineator a comment to that effect (here and throughout).

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+            The ISR processed exposure to me measured.
+        camera : `lsst.afw.cameraGeom.Camera`
+             The camera geometry for this exposure.
+
+        Returns
+        -------
+        result : `lsst.pipe.base.Struct`
+            Result struct with components:
+            - ``outputStats`` : `dict`
+                The output measured statistics.
```
laurenam, line 194: Is specifying "output" really necessary here?  (i.e. were there input stats you are differentiating from?)

Are you making the return a struct (for a single element) in anticipation of adding to it in the future?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        return ampStats
+
+    def unmaskedImageStats(self, exposure, statControl):
+        """Measure amplifier level statistics on the data read from disk,
+        ignoring all mask planes.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure to measure.
```
laurenam, line 378: "to measure" is not the most descriptive...

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        statControl : `lsst.afw.math.StatControl`
+            Statistics control object with the parameters defined by
+            the config.
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, scalar]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the statistics measured and their values.
+
```
laurenam: Blank line not needed.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
```
laurenam, line 408: Blank line not needed.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        scaledExposure = exposure.clone()
+        exposureTime = scaledExposure.getInfo().getVisitInfo().getExposureTime()
```
laurenam, line 415: Is this guaranteed to give you a finite value?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+    detectorStatKeywords = pexConfig.DictField(
+        keytype=str,
+        itemtype=str,
+        doc="Detector statistics to create.",
+        default={},
+    )
+
+
+class CpVerifyStatsTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
+    """
```
laurenam: Fill me in please!

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+            Statistics control object with the parameters defined by
+            the config.
+
+        Returns
+        -------
+        ampStats : `dict` [`str`, `dict` [`str`, scalar]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the statistics measured and their values.
+        """
+
```
laurenam: Blank line not needed.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        if len(self.config.normImageStatKeywords):
+            outputStats = mergeStatDict(outputStats, self.normalizedImageStats(exposure, statControl))
+
+        if len(self.config.crImageStatKeywords):
+            outputStats = mergeStatDict(outputStats, self.crImageStats(exposure, statControl))
+
+        return outputStats
+
+    def amplifierStats(self, exposure, keywordDict, statsControl):
+        """Measure amplifier level statistics on the data read from disk.
```
laurenam: I don't really get the "read from disk" specification....

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+            for k, v in statAccessor.items():
+                theseStats[k] = stats.getValue(v)
+            ampStats[ampName] = theseStats
+
+        return ampStats
+
+    def unmaskedImageStats(self, exposure, statControl):
+        """Measure amplifier level statistics on the data read from disk,
+        ignoring all mask planes.
```
laurenam: I think it's often (incorrectly) assumed that the "ignoring all mask planes" is to be interpreted as "ignoring all pixels with any mask plane set".  I'm definitely not insisting, but it may be worthwhile to rephrase along the lines of "including all pixels in the exposure, regardless of any mask planes set".

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        doc="Repair task to use.",
+    )
+    psfFwhm = pexConfig.Field(
+        dtype=float,
+        default=3.0,
+        doc="Repair PSF FWHM (pixels).",
+    )
+    psfSize = pexConfig.Field(
+        dtype=int,
+        default=21,
```
laurenam, line 99: Do we have any justification/mandate for the defaults?  Might the docstring provide some guidance as to if/when one might want to have different settings?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        doc="Repair PSF size (pixels).",
+    )
+    crGrow = pexConfig.Field(
+        dtype=int,
+        default=2,
```
laurenam, line 104: Again, do we have any justification/mandate for this default?

setup.cfg
taranu: Not to pile onto another reviewer's responsibility again but this looks like an old `setup.cfg` that's missing a lot of things from the current [project template](https://github.com/lsst/templates/blob/master/project_templates/stack_package/example/setup.cfg). I'm not sure if DM-14207 is relevant anymore even if it's still to-do?

python/lsst/__init__.py
czwa: I've removed all of the `lsstimport` imports in these files.

python/lsst/cp/verify/__init__.py
czwa: I've removed those lines.

I've also added the boilerplate at the start of all the `__init__.py` files.

python/lsst/cp/verify/utils.py
czwa: This is a "thing that should never happen", so I've added a check that will raise in the loop if a value would be over-written.

python/lsst/cp/verify/verifyBias.py
czwa: I kept it as one key-value pair per line to increase legibility.  I find it easier to scan down a list to ensure everything is included than to parse it from a block.

python/lsst/cp/verify/verifyBias.py
czwa: These should be updated, and I've updated the docstring for the main `CpVerifyStatsTask`.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+    normImageStatKeywords = pexConfig.DictField(
+        keytype=str,
+        itemtype=str,
+        doc="Image statistics to run on expTime normalized amplifier segments.",
+        default={},
+    )
+    catalogStatKeywords = pexConfig.DictField(
+        keytype=str,
+        itemtype=str,
+        doc="Image statistics to run on expTime normalized amplifier segments.",
```
laurenam: You stuttered with the docstring for `normImageStatKeywords` above.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        default={},
+    )
+    detectorStatKeywords = pexConfig.DictField(
+        keytype=str,
+        itemtype=str,
+        doc="Detector statistics to create.",
```
laurenam: Doc could use a tad more detail...

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        self.makeSubtask("repair")
+
+    def run(self, inputExp, camera):
+        """Calculate quality statistics and verify they meet the requirements
+        for a calibration.
+
+        Parameters
+        ----------
+        inputExp : `lsst.afw.image.Exposure`
+            The ISR processed exposure to me measured.
```
laurenam: to me, or not to me...

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        camera : `lsst.afw.cameraGeom.Camera`
+             The camera geometry for this exposure.
```
laurenam: "The camera geometry for \`\`inputExp\`\`."?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+        outputStats = self._emptyAmpDict(exposure)
+
+        if len(self.config.imageStatKeywords):
+            outputStats = mergeStatDict(outputStats, self.amplifierStats(exposure,
+                                                                         self.config.imageStatKeywords,
+                                                                         statControl))
+
+        if len(self.config.unmaskedImageStatKeywords):
+            outputStats = mergeStatDict(outputStats, self.unmaskedImageStats(exposure, statControl))
```
laurenam: Why pass the `statsControl` for this one?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+        return outputStats
+
+    def amplifierStats(self, exposure, keywordDict, statsControl):
+        """Measure amplifier level statistics on the data read from disk.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure to measure.
```
laurenam, line 332: Thin doc...

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        """Measure amplifier level statistics on the data read from disk,
+        ignoring all mask planes.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure to measure.
+        statControl : `lsst.afw.math.StatControl`
+            Statistics control object with the parameters defined by
+            the config.
```
laurenam: Remove here too.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+            dictionaries of the statistics measured and their values.
+
+        """
+        noMaskStatsControl = afwMath.StatisticsControl(self.config.numSigmaClip,
+                                                       self.config.clipMaxIter,
+                                                       0x0)
+        return self.amplifierStats(exposure, self.config.unmaskedImageStatKeywords, noMaskStatsControl)
+
+    def normalizedImageStats(self, exposure, statControl):
+        """Measure amplifier level statistics on the data read from disk,
```
laurenam: Again, why specify "read from disk"?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        self.repair.run(crRejectedExp, keepCRs=False)
+        if self.config.crGrow > 0:
+            crMask = crRejectedExp.getMaskedImage().getMask().getPlaneBitMask("CR")
+            spans = afwGeom.SpanSet.fromMask(crRejectedExp.mask, crMask)
+            spans = spans.dilated(self.config.crGrow)
+            spans.setMask(crRejectedExp.mask, crMask)
+
+        return self.amplifierStats(crRejectedExp, self.config.crImageStatKeywords, statControl)
+
+    # Methods that need to be implemented by the calibration-level subclasses.
```
laurenam, line 456: I don't see any methods that actually define `catalogStatisitics()` or `detectorStatistics()` functions.  Is this in anticipation of yet-to-be-added calibration-level subclasses?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, scalar]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the statistics measured and their values.
+        """
+        raise NotImplementedError("Subclasses must implement verification criteria.")
+
+    def detectorStatistics(self, statisticsDictionary, statControl):
+        """Calculate detector level statistics based on the existing measurements.
```
laurenam: Could you elaborate on "existing measurements"?  Is this meant to aggregate pre-measured per-`amp` statistics into per-`detector` statistics?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+        Parameters
+        ----------
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
```
laurenam, line 482: Superfluous comma?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+            Statistics control object with the parameters defined by
+            the config.
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, `dict` [`str`, scalar]]
+            A dictionary indexed by the amplifier name, containing
+            dictionaries of the statistics measured and their values.
+        """
+        raise NotImplementedError("Subclasses must implement verification criteria.")
```
laurenam: I don't get the "verification" component implied by this function whose definition states that is simple calculates a set of statistics.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+        Parameters
+        ----------
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
+            Dictionary of measured statistics.
+
+        Returns
+        -------
+        outputStats : `dict` [`str`, scalar]
+            A dictionary of the statistics measured and their values.
```
laurenam, line 491: Please elaborate on that the key value pairs are and if this really is an aggregated set of the measurements in `statisticsDictionary`.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+        Raises
+        ------
+        NotImplementedError :
+            This method must be implemented by the calibration-type
+            subclass.
+        """
+        raise NotImplementedError("Subclasses must implement verification criteria.")
```
laurenam: Again, what is the "verification" component here?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+    def verify(self, exposure, statisticsDictionary):
+        """Verify if the measured statistics meet the verification criteria.
```
laurenam: whether (or not)?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        success : `bool`
+            A boolean indicating if all tests have passed.
+
+        Raises
+        ------
+        NotImplementedError :
+            This method must be implemented by the calibration-type
+            subclass.
+        """
+        raise NotImplementedError("Subclasses must implement verification criteria.")
```
laurenam, line 528: I get it here 😃

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,332 @@
+#
```
laurenam: I don't think this initial empty comment line is necessary, e.g. https://github.com/lsst/templates/blob/master/file_templates/stack_license_preamble_py/example.py

python/lsst/cp/verify/utils.py
```diff
@@ -0,0 +1,44 @@
+#
```
laurenam: No need for initial empty comment line.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -0,0 +1,91 @@
+#
```
laurenam: No need for initial empty comment line.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
+#
```
laurenam: No need for initial empty comment line.

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MEAN'], 2.08672, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['NOISE'], 13.99547, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['CR_NOISE'], 14.11526, 4)
+
+        self.assertIn(resultStats['SUCCESS'], [True, False])
+
+
+class VerifyDarkTestCase(lsst.utils.tests.TestCase):
+    """Unit test for stats code.
```
laurenam: Another stutter...

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MEAN'], 2.0043, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['NOISE'], 3.12948, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['CR_NOISE'], 3.15946, 4)
+
+        self.assertIn(resultStats['SUCCESS'], [True, False])
+
+
+class VerifyDefectsTestCase(lsst.utils.tests.TestCase):
+    """Unit test for stats code.
```
laurenam: St-st-stutter. Slightly more detailed summaries, please!

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        self.inputExp.getImage().getArray()[40, 50] = 100000
+        self.inputExp.getImage().getArray()[75, 50] = np.nan
+
+        makeMockExp(self.inputExp, addCR=False)
+
+        self.inputExp.getMask().getArray()[0, 0] = 1
+        self.inputExp.getMask().getArray()[40, 50] = 1
+        self.inputExp.getMask().getArray()[75, 50] = 1
+
+        self.inputExp.writeFits("/home/czw/tmp/fake_flat.fits")
```
laurenam: Well...that's not gonna work 😉

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+import unittest
+
+import lsst.utils.tests
+import lsst.ip.isr.isrMock as isrMock
+import lsst.cp.verify as cpVerify
+import lsst.ip.isr.isrFunctions as isrFunctions
+
+
+def makeMockExp(exposure, addCR=True):
+    """Update an exposure with a mask and variance plane.
```
laurenam, line 32: So really this is `updateMockExp`?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Parameters
+    ----------
+    exposure : `lsst.afw.image.Exposure`
+        Exposure to be modified in place.
+    addCR ; `bool`
+        Whether the cosmic ray should be added.
```
laurenam: "Whether a single cosmic ray should be inserted into \`\`exposure\`\`."?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+
+    def setUp(self):
+        # Generate a mock exposure/camera to test.
+        self.inputExp = isrMock.CalibratedRawMock().run()
+        self.camera = isrMock.IsrMock().getCamera()
+
+        makeMockExp(self.inputExp)
+
+    def test_failures(self):
```
laurenam, line 75: This is the first snake_case I'm seeing in this repo...since this is a new package, I invite you to peruse https://developer.lsst.io/python/style.html?highlight=camelcase#naming-conventions-for-science-pipelines and decide what you think is appropriate throughout.

ups/cp_verify.cfg
```diff
@@ -0,0 +1,13 @@
+# -*- python -*-
+
+import lsst.sconsUtils
+
+dependencies = {
+    "required": ["ip_isr", "cp_pipe"],
+    "optional": ["sdqa"],
+}
```
laurenam, line 16: I'm not familiar with the details of this file and what it needs to entail, but I note that the list of "required" dependencies is much shorter than that in `ups/cp_verify.table`.  And I don't see why `sdqa` is needed (even as optional...nor do I pretend to have any idea what it is, so I may be missing something! Oh, looks like it's legacy: https://github.com/lsst-dm/legacy-sdqa)?

The template with some guidelines is here: https://github.com/lsst/templates/blob/master/project_templates/stack_package/example_subpackage/ups/example_subpackage.cfg

ups/cp_verify.table
```diff
@@ -0,0 +1,15 @@
...
+setupRequired(afw)
+setupRequired(geom)
+setupRequired(utils)
+setupRequired(pex_exceptions)
+
+envPrepend(DYLD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(LSST_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
+envPrepend(PYTHONPATH, ${PRODUCT_DIR}/python)
+envPrepend(PATH, ${PRODUCT_DIR}/bin)
```
laurenam, line 16: Is this one required (I don't think you get a `bin/` directory on building this package)?

ups/cp_verify.table
```diff
@@ -0,0 +1,15 @@
+setupRequired(cp_pipe)
+setupRequired(ip_isr)
+setupRequired(meas_algorithms)
+setupRequired(pipe_base)
+setupRequired(sconsUtils)
+setupRequired(afw)
+setupRequired(geom)
```
laurenam: I don't see `geom` imported anywhere.

ups/cp_verify.table
```diff
@@ -0,0 +1,15 @@
...
+setupRequired(utils)
+setupRequired(pex_exceptions)
```
laurenam: Need a few more? 
E.g.
`pex_config`: https://github.com/lsst/cp_verify/blob/tickets/DM-27485/python/lsst/cp/verify/verifyStats.py#L28
`pipe_tasks`: https://github.com/lsst/cp_verify/blob/tickets/DM-27485/python/lsst/cp/verify/verifyStats.py#L33

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def makeMockExp(exposure, addCR=True):
+    """Update an exposure with a mask and variance plane.
+
+    Parameters
+    ----------
+    exposure : `lsst.afw.image.Exposure`
+        Exposure to be modified in place.
+    addCR ; `bool`
```
laurenam: Colon (not semicolon)

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+            # Or the detector stats
+            config.catalogStatKeywords = {}
+            config.detectorStatKeywords = {'DET_SIGMA', 'STDEV'}
+            task = cpVerify.CpVerifyStatsTask(config=config)
+            results = task.run(self.inputExp, self.camera)
+        self.assertIsNone(results)
+
+    def test_generic(self):
+        # Test a subset of the output values to identify that the
+        # image stat methods haven't changed.
```
laurenam: Why not make this a formal doc summary (i.e. with """s)?  Here and throughout...

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+                                        'norm_SIGMA': 'STDEV'}
+        task = ToySubClass(config=config)
+
+        results = task.run(self.inputExp, self.camera)
+        resultStats = results.outputStats
+
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MEAN'], 1506.06976, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['un_MEAN'], 1501.0299, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['norm_MEAN'], 301.213957, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['cr_MEAN'], 1504.2776, 4)
```
laurenam, line 121: Do you have a justification for the tolerances?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self.assertTrue(resultStats['VERIFY']['A REAL TEST'])
+        self.assertFalse(resultStats['VERIFY']['A BAD TEST'])
+
+        self.assertTrue(resultStats['SUCCESS'])
+
+
+class VerifyBiasTestCase(lsst.utils.tests.TestCase):
+    """Unit test for stats code.
```
laurenam: Stutter from previous class summary.

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        makeMockExp(self.inputExp)
+
+        self.camera = isrMock.IsrMock().getCamera()
+
+    def test_bias(self):
+        # Test a subset of the output values to identify that the
+        # image stat methods haven't changed.
+        task = cpVerify.CpVerifyBiasTask()
+        results = task.run(self.inputExp, self.camera)
+        resultStats = results.outputStats
```
laurenam: Might be helpful to name these, e.g. `biasResults` & `biasStats`, so it's clear which set we're looking at in this function (and same for other calib-level tests below).

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MEAN'], 2.08672, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['NOISE'], 13.99547, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['CR_NOISE'], 14.11526, 4)
+
+        self.assertIn(resultStats['SUCCESS'], [True, False])
```
laurenam: Shouldn't you be able to test which it actually is (success or not)?  Oh, maybe you wan't to leave that open as thresholds for pass/fail might change and that need not break this test?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        config.doAddFlat = False
+        config.doAddFringe = False
+
+        config.skyLevel = 1000
+        config.rngSeed = 12345
+        self.inputExp = isrMock.IsrMock(config=config).run()
+
+        # These are simulated defects
+        self.inputExp.getImage().getArray()[0, 0] = -100000
+        self.inputExp.getImage().getArray()[40, 50] = 100000
```
laurenam: Since you are testing against these numbers below, perhaps define a variable?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        self.inputExp.getImage().getArray()[75, 50] = np.nan
+
+        makeMockExp(self.inputExp, addCR=False)
+
+        self.inputExp.getMask().getArray()[0, 0] = 1
+        self.inputExp.getMask().getArray()[40, 50] = 1
+        self.inputExp.getMask().getArray()[75, 50] = 1
```
laurenam, line 237: Similarly, make defect positions variables?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        self.camera = isrMock.IsrMock().getCamera()
+
+    def test_defects(self):
+        # Test a subset of the output values to identify that the
+        # image stat methods haven't changed.
+        task = cpVerify.CpVerifyDefectsTask()
+        results = task.run(self.inputExp, self.camera)
+        resultStats = results.outputStats
+
+        self.assertEqual(resultStats['AMP']['C:0,0']['DEFECT_PIXELS'], 53)
```
laurenam: Are you comfortable only testing the values of the first `amp`?

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        self.assertEqual(resultStats['AMP']['C:0,0']['OUTLIERS'], 17)
+        self.assertEqual(resultStats['AMP']['C:0,0']['STAT_OUTLIERS'], 3)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MEDIAN'], 999.466, 4)
```
laurenam: Seems odd to require equivalence to 4 decimal places when only 3 are specified (I guess 0s are assumed!)

tests/test_verifyStats.py
```diff
@@ -0,0 +1,264 @@
...
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['STDEV'], 30.96303, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MIN'], 881.56146, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['MAX'], 1124.19934, 4)
+
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['UNMASKED_MIN'], -100000, 4)
+        self.assertAlmostEqual(resultStats['AMP']['C:0,0']['UNMASKED_MAX'], 100000, 4)
```
laurenam: What is the reason these can't be exact (e.g. some kind of background adjustment after they are added?)

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+
+    def normalizedImageStats(self, exposure, statControl):
+        """Measure amplifier level statistics on the data read from disk,
+        after dividing by the exposure time.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure to measure.
+        statControl : `lsst.afw.math.StatControl`
```
laurenam: `lsst.afw.math.StatisticsControl`

And it might be worth just calling the parameter `statisticsControl` for consistency?

python/lsst/cp/verify/verifyStats.py
```diff
@@ -0,0 +1,510 @@
...
+        if detector is None:
+            raise RuntimeError("No detector found in exposure!")
+
+        for amp in detector.getAmplifiers():
+            outputStats[amp.getName()] = {}
+
+        return outputStats
+
+    # Image measurement methods.
+    def imageStatistics(self, exposure, statControl):
```
laurenam, line 286: You lost the second "s" on `statsControl` from above (and I still think it might be easier to just call them all `statisticsControl`...but I only insist on consistency 😉 ).

python/lsst/cp/verify/verifyBias.py
czwa: I've changed all internal variables to be `outputStatistics`.  This will disambiguate them from the `outputStats` dataset that the output product.  I've also changed all `*Dictionary` variables to `*Dict` to match how the utility functions are named.

python/lsst/cp/verify/verifyBias.py
czwa: These are the test numbers as defined by DMTN-101.

python/lsst/cp/verify/utils.py
czwa: These dictionaries are created by the `CpVerifyStatsTask._emptyAmpDict(exposure)` method, which should remove the need to pass the detector along with the two dictionaries to ensure the amps match.

.gitignore
czwa: Added.

README.md
czwa: I've added a ticket that will contain how-to documentation.  I'd like to get the calibration coverage improved, as I suspect there will be unique cases in some of the other calibration types (crosstalk and brighter-fatter seem the most troublesome right now).  
https://jira.lsstcorp.org/browse/DM-29798

setup.cfg
czwa: Updated to match the current template.

python/lsst/cp/verify/verifyStats.py
czwa: The test path is to generate the proposed calibration, and then apply ISR to the input exposures that were used to generate the calibration, using the calibration as the appropriate bias, dark, etc.  However, a number of the DMTN-101 tests suggest comparing clipped statistics on these postISR exposures to "straight" statistics on the CR cleaned postISR images.  If those are consistent, then the outliers in the image distribution are from the input exposures, and not bad pixels in the calibration.

python/lsst/cp/verify/verifyStats.py
czwa: I've updated it to reflect that it's the size of the bounding box.

python/lsst/cp/verify/verifyStats.py
czwa: Reworded for clarity.

python/lsst/cp/verify/verifyStats.py
czwa: Gen3 tasks must return a struct containing with keys that match the output connection objects.  I named it `outputStats` in the connection class to note that it's the output (and I find it easier to visualize the pipeline steps when I call them `inputX` and `outputY`).  I then set the dataset_type_name to `detectorStats`, so the butler object has a clear name.

python/lsst/cp/verify/verifyStats.py
czwa: In a perfect world.  I've added a raise to guard against reality.

python/lsst/cp/verify/verifyStats.py
czwa: These are the defaults inherited (via `cp_pipe`) from `pipe_drivers` `constructCalibs.py`.  This is run on bias, dark, and flat images, none of which have any real sources that would inform PSF options.  Thinking through the numbers, this should yield a narrow PSF (`\sigma ~ 1.27`) reaching out to approximately 8 sigma with this PSF size.  I don't think I have any strong guidance if it didn't remove all the obvious CRs, other than changing the PSF FWHM while keeping the size appropriate to prevent truncation.

python/lsst/cp/verify/verifyStats.py
czwa: Another inherited value.  This would also leave noticeable residuals (outlines where CRs were) if it were significantly wrong.

python/lsst/cp/verify/verifyStats.py
czwa: I was originally going to modify the object instead of making a new one.  I've removed it.

python/lsst/cp/verify/verifyStats.py
czwa: Yes, crosstalk and brighter-fatter will likely need to do catalog level work, and flats will need detector level statistics.

python/lsst/cp/verify/verifyStats.py
czwa: Yes, I don't currently think that anything needs to be measured detector-wide, so this should just be some sort of summation of the per-amplifier stats.  As an example, flat verification will check that the variance of the amplifier means is small (tbd what that means), indicating that the flat incorporates any gain differences between amps.

ups/cp_verify.cfg
czwa: ~Following the python-only template, I've removed this file, as it shouldn't be needed.~
I've readded an updated version to fix the package not building.

ups/cp_verify.table
czwa: I'm going to leave it in for now, as I had hoped to get `bin.src` scripts written, but have had to push those until a later ticket (`pipetask` can't deal with multiple calibration repos, so there will need to be individual scripts to directly compare calibrations).

ups/cp_verify.table
czwa: I've also alphabetized them, as they were before I added more packages.

python/lsst/cp/verify/verifyStats.py
czwa: I've updated the tests to use the 3.0 value explicitly, as the initial target values were defined based on that clipping.  This prevents needing to skew the distribution to get 5.0 events in the tests.

tests/test_verifyStats.py
czwa: They're set so the values listed pass, without having to carry around too many extra decimal places that don't yield much value.  As a number of these tests failed when the default `numSigmaClip=5.0` was run, they do perform the goal I had for them: ensure changes to the algorithm are caught.

tests/test_verifyStats.py
czwa: Yeah, I just want the test to check that no test is inserting a non-boolean where we expect a boolean.

tests/test_verifyStats.py
czwa: I've added a class variable so the generation and testing match.

tests/test_verifyStats.py
czwa: I have not done this, as these are only used in the `setUp` method.

tests/test_verifyStats.py
czwa: Expanded value is `999.4660034179688`.

tests/test_verifyStats.py
czwa: For now, I am.  The defects verification probably needs more thought, and that may change this test.

tests/test_verifyStats.py
czwa: No.  Switched to `.assertEqual`.

python/lsst/cp/verify/verifyBias.py
laurenam: Worth adding that to the comment?

python/lsst/cp/verify/utils.py
laurenam: Sure they are...there, but this function lives in a general `utils.py` file, so some specifics about assumptions/requirements of the input parameters seems warranted.

python/lsst/cp/verify/verifyStats.py
laurenam: I'm not sure I've fully wrapped my head around that...but ok!

python/lsst/cp/verify/verifyStats.py
laurenam: I'd go one step further and update the variable name (a naked "size" is just too ambiguous).  Your call, though!

python/lsst/cp/verify/verifyStats.py
laurenam: Not totally clear on this point either (are you saying this is consistent with the DMTN-101 requirements, or does that document need updating?), but ok!

tests/test_verifyStats.py
laurenam: Oh, yeah, but they are used twice, so could conceivably get out of sync...totally not insisting though!

python/lsst/cp/verify/verifyBias.py
czwa: Added.

python/lsst/cp/verify/utils.py
czwa: Added raise for this case.

python/lsst/cp/verify/verifyStats.py
czwa: Sorry, I was unclear (the problem of working on these comments while doing too many other things).  The default is now set to 5.0 here in verifyStats.py.  When running this for the unit tests, I'm explicitly switching back to the 3.0 value.  This allows the test images to be a reasonable size while still having normally distributed errors that get clipped values different than the unclipped values.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,330 @@
...
+        name="exposureStats",
+        doc="Output statistics.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "exposure"],
+    )
+
+
+class CpVerifyExpMergeConfig(pipeBase.PipelineTaskConfig,
+                             pipelineConnections=CpVerifyExpMergeConnections):
+    """Configuration paramters for exposure stats merging.
```
taranu: parameters

python/lsst/cp/verify/mergeResults.py
```diff
@@ -0,0 +1,330 @@
...
+        outputs = self.run(**inputs)
+        butlerQC.put(outputs, outputRefs)
+
+    def run(self, inputStats, camera, inputDims):
+        """Merge statistics.
+
+        Parameters
+        ----------
+        inputStats : `list` [`dict`]
+            Measured statistics for a detector.
```
taranu: I realize that you'll address this in the future but for now, is the inputStats dict the same structure as for outputStats? Note briefly if so (and definitely point to what it is if not).

tests/test_mergeResults.py
```diff
@@ -0,0 +1,144 @@
...
+                ampDict[ampName][value] = np.random.normal(10.0, 1.2)
+
+        ampVerify = dict()
+        detVerify = dict()
+        catVerify = dict()
+        for ampName in ampDict:
+            ampVerify[ampName] = dict()
+            ampSuccess = True
+            for value in valueNames:
+                if ampDict[ampName][value] < 14.0:
```
taranu: Should be `< threshold`, no?

taranu (commented):
Found a few minor issues to fix.

PR #3, opened 2021-07-27
DM-31146: Add cp_verify to lsst_distrib

czwa:
This incorporates changes from the u/ktl/fix-build branch.  As suggested, this fixes the build.


timj (approved):
Looks okay. There is a merge commit on this branch that looks like it needs to be rebased out of existence.

PR #5, opened 2021-08-25
DM-30169: Define flat tests for cp_verify

czwa:
This defines an initial set of flat verification tests:

1. Per amplfier noise should be consistent with a Poissonian distribution.
2. Amplifiers on one detector should have the same mean level.
3. Detectors in an exposure should have the same mean level.

Additional helper code is added, as well as the ability to measure per-exposure statistics.

### DIFF
--- a/pipelines/VerifyDefectPostFlat.yaml
+++ b/pipelines/VerifyDefectPostFlat.yaml
@@ -1,0 +1,41 @@
+description: cp_verify DEFECT calibration verification
+tasks:
+  verifyDefectApply:
+    class: lsst.ip.isr.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'verifyDefectProc'
+      overscan.fitType: 'MEDIAN_PER_ROW'
+      doWrite: True
+      doOverscan: True
+      doAssembleCcd: True
+      doBias: True
+      doVariance: False
+      doLinearize: False
+      doCrosstalk: False
+      doBrighterFatter: False
+      doDark: True
+      doStrayLight: False
+      doFlat: True
+      doFringe: False
+      doApplyGains: False
+      doDefect: True
+      doSaturationInterpolation: False
+      growSaturationFootprintSize: 0
+  verifyMeasureStats:
+    class: lsst.cp.verify.CpVerifyDefectsTask
+    config:
+      connections.inputExp: 'verifyDefectProc'
+      connections.outputStats: 'verifyDefectDetStats'
+      doVignette: False
+      useReadNoise: False
+  verifyExposureStats:
+    class: lsst.cp.verify.CpVerifyExpMergeTask
+    config:
+      connections.inputStats: 'verifyDefectDetStats'
+      connections.outputStats: 'verifyDefectExpStats'
+  verifyRunStats:
+    class: lsst.cp.verify.CpVerifyRunMergeTask
+    config:
+      connections.inputStats: 'verifyDefectExpStats'
+      connections.outputStats: 'verifyDefectStats'
--- a/pipelines/VerifyDefectPostFlat.yaml
+++ b/pipelines/VerifyDefectPostFlat.yaml
@@ -6,29 +6,29 @@ tasks:
       connections.ccdExposure: 'raw'
       connections.outputExposure: 'verifyDefectProc'
       overscan.fitType: 'MEDIAN_PER_ROW'
-      doWrite: True
-      doOverscan: True
-      doAssembleCcd: True
-      doBias: True
-      doVariance: False
-      doLinearize: False
-      doCrosstalk: False
-      doBrighterFatter: False
-      doDark: True
-      doStrayLight: False
-      doFlat: True
-      doFringe: False
-      doApplyGains: False
-      doDefect: True
-      doSaturationInterpolation: False
+      doWrite: true
+      doOverscan: true
+      doAssembleCcd: true
+      doBias: true
+      doVariance: false
+      doLinearize: false
+      doCrosstalk: false
+      doBrighterFatter: false
+      doDark: true
+      doStrayLight: false
+      doFlat: true
+      doFringe: false
+      doApplyGains: false
+      doDefect: true
+      doSaturationInterpolation: false
       growSaturationFootprintSize: 0
   verifyMeasureStats:
     class: lsst.cp.verify.CpVerifyDefectsTask
     config:
       connections.inputExp: 'verifyDefectProc'
       connections.outputStats: 'verifyDefectDetStats'
-      doVignette: False
-      useReadNoise: False
+      doVignette: false
+      useReadNoise: false
   verifyExposureStats:
     class: lsst.cp.verify.CpVerifyExpMergeTask
     config:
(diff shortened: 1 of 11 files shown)


pipelines/VerifyDefectPostFlat.yaml
```diff
@@ -0,0 +1,41 @@
+description: cp_verify DEFECT calibration verification
+tasks:
+  verifyDefectApply:
+    class: lsst.ip.isr.IsrTask
+    config:
+      connections.ccdExposure: 'raw'
+      connections.outputExposure: 'verifyDefectProc'
+      overscan.fitType: 'MEDIAN_PER_ROW'
+      doWrite: True
```
plazas: I believe once Nate told me that booleans in the YAML scripts should start with lowercase. I'm not sure if there's anything official in the guide (I think not), and I might have seen both conventions, so I'll leave it at your discretion :)

python/lsst/cp/verify/mergeResults.py
```diff
@@ -128,11 +128,14 @@ def run(self, inputStats, camera, inputDims):
         outputStats = {}
         success = True
 
+        mergedStats = {}
```
plazas, line 131: Are you merging the metrics/stats from all detectors to do an overall statistics test?  Or what is the purpose of the merging?

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -0,0 +1,199 @@
...
+        success : `bool`
+            A boolean indicating if all tests have passed.
+        """
+        ampStats = statisticsDict['AMP']
+        verifyStats = {}
+        success = True
+        for ampName, stats in ampStats.items():
+            verify = {}
+
+            # DMTN-101 Test 10.X: confirm that per-amplifier scatter is
```
plazas, line 108: Am I looking at the wrong version of the DMTN-101 document? https://dmtn-101.lsst.io/   I can't see the definition of the tests that you implement here under section 10.

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -0,0 +1,199 @@
...
+        return outputStatistics
+
+    def verify(self, exposure, statisticsDict):
+        """Verify that the measured statistics meet the verification criteria.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure the statistics are from.
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
```
plazas: The name ``statisticsDictionary`` is different from ``statisticsDict`` in the argument.

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -0,0 +1,199 @@
...
+            if verify['SUCCESS'] is False:
+                success = False
+
+            verifyStats[ampName] = verify
+
+        verifyDet = {}
+        detStats = statisticsDict['DET']
+
+        # DMTN-101 Test 10.Y: confirm intra-chip scatter is small.
+        verifyDet['SCATTER'] = bool(detStats['SCATTER']/detStats['MEAN'] <= 0.05)
```
plazas, line 122: This definition of "small" (0.05) is not expected to be modified? Just pointing that out in case that we want to make it a config parameter.

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -0,0 +1,199 @@
...
+    def verify(self, exposure, statisticsDict):
+        """Verify that the measured statistics meet the verification criteria.
+
+        Parameters
+        ----------
+        exposure : `lsst.afw.image.Exposure`
+            The exposure the statistics are from.
+        statisticsDictionary : `dict` [`str`, `dict` [`str`, scalar]],
+            Dictionary of measured statistics.  The inner dictionary
+            should have keys that are statistic names (`str`) with
```
plazas, line 90: Maybe put some examples of the expected statistics names, e.g., "SCATTER", "MEAN", etc.?

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -0,0 +1,199 @@
...
+        -------
+        outputStatistics : `dict` [`str, scalar]
+            A dictionary of the statistics measured and their values.
+        """
+        detectorMeans = []
+        for detName, stats in statisticsDictionary.items():
+            # Get detector stats:
+            detectorMeans.append(stats['DET']['MEAN'])
+
+        return {'SCATTER': np.stdev(detectorMeans)}
```
plazas, line 167: why do we use NumPy and not afw? It doesn't matter?

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -0,0 +1,199 @@
...
+            dictionaries of the verification criteria.
+        success : `bool`
+            A boolean indicating if all tests have passed.
+
+        """
+        verifyStats = {}
+        success = True
+
+        # DMTN-101 Test 10.Z: confirm inter-chip scatter is small.
+        verifyStats['SCATTER'] = bool(statisticsDictionary['EXP']['SCATTER'] <= 0.05)
```
plazas, line 195: Same comment as above about hard-coded 0.05.

plazas (approved):
I think it looks good; I just left a few comments regarding clarification.

pipelines/VerifyDefectPostFlat.yaml
timj: Yes, if you run yamllint the default is to prefer `true` for this. Consider adding the yamllint GitHub Action.

czwa: I've added yamllint, pushed, and after fixing things up, it seems to be working.  I'll file a ticket to do the same for cp_pipe, as those pipelines were also written by me, and almost certainly have capital letters.

python/lsst/cp/verify/mergeResults.py
czwa: The goal of the merge is to allow there to be a hierarchical path from a single top level file (the output of `CpVerifyRunMergeTask`) that contains a list of all the failures, through the exposure level down to the detector/amp level.  The visualization notebooks (which I still need to extend for multi-detector cameras) can start with the top level file, and use that to determine what needs to be examined because of the failure.

python/lsst/cp/verify/verifyFlat.py
czwa: 10.X was my shorthand for "this will be defined in section 10, but isn't formally done yet."  I want to get the code written first, because adding/updating tests is easy once the code exists, and getting something running is easier than getting DMTN-101 updates approved.

python/lsst/cp/verify/verifyFlat.py
czwa: Fixed here and elsewhere.

python/lsst/cp/verify/verifyFlat.py
czwa: I've been working with the assumption that once we finish DMTN-101, the tests will be fixed; those will be the criteria we use for calibrations.

python/lsst/cp/verify/verifyFlat.py
czwa: I didn't think there'd be any difference.  This is a straight standard deviation, so they should be algorithmically identical (no clipping or masking needed).

python/lsst/cp/verify/verifyFlat.py
czwa: The names are somewhat arbitrary.  I didn't want to force the name to equal the afwStatistics string name, as I know there will be metrics that are not calculated from afwStatistics (brighter-fatter will have a size-flux slope as an example).

PR #7, opened 2021-09-08
DM-31692: Fix edge handling in cp_verify CR cases

czwa:
Grown CR spans should fit on the source image; clip them to be sure.

### DIFF
--- a/python/lsst/cp/verify/verifyStats.py
+++ b/python/lsst/cp/verify/verifyStats.py
@@ -472,6 +472,7 @@ class CpVerifyStatsTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
             crMask = crRejectedExp.getMaskedImage().getMask().getPlaneBitMask("CR")
             spans = afwGeom.SpanSet.fromMask(crRejectedExp.mask, crMask)
             spans = spans.dilated(self.config.crGrow)
+            spans = spans.clippedTo(crRejectedExp.getBBox())
             spans.setMask(crRejectedExp.mask, crMask)
 
         return self.amplifierStats(crRejectedExp, self.config.crImageStatKeywords, statControl)


python/lsst/cp/verify/verifyStats.py
```diff
@@ -472,6 +472,7 @@ def crImageStats(self, exposure, statControl):
             crMask = crRejectedExp.getMaskedImage().getMask().getPlaneBitMask("CR")
             spans = afwGeom.SpanSet.fromMask(crRejectedExp.mask, crMask)
             spans = spans.dilated(self.config.crGrow)
+            spans = spans.clippedTo(crRejectedExp.getBBox())
```
plazas, line 475: just to make sure o what is happening here, after the CR span is grown, it could have done so because it included an edge pixel, which of course should not belong to the CR span. is that correct? What this function does is to clip that span to the size of the CRs span, right?

plazas (approved):
Thanks for fixing this. I also tested at the summit with LATISS data, and it seemed to have worked.  I just asked a question to clarify what the function in the one-liner fix is doing. Along those lines, I'll leave it to your consideration whether the new addition and maybe the surrounding code warrant an explanatory comment (for others, and for our future selves ;)).

python/lsst/cp/verify/verifyStats.py
czwa: It clips the grown set of spans to the bounding box of the image.  The error was arising because a CR located at (0, 0) grown isotropically by one pixel would include {(-1, -1), (-1, 0), (0, -1)}.  Those all lie outside the image, so applying that as a mask was triggering the error.  I think this is formally restricting the spanset to the intersection of the initial set (L474) and the image bbox.

PR #8, opened 2021-11-10
DM-30172: Define BFK tests for cp_verify

czwa:
This adds catalog support to cp_verify, and then uses that to compare two ISR runs: one with and one without brighter-fatter correction.  If the sizes trend in the correct direction, then the correction has improved things and is likely good.


examples/cpVerifyBfk.ipynb
```diff
@@ -0,0 +1,763 @@
+{
```
mfisherlevine, line 1: Loving the inclusion of the notebook, looks great. Would you like an actual review on this too?

pipelines/VerifyBfk.yaml
```diff
@@ -0,0 +1,49 @@
+description: cp_verify BFK calibration verification
```
mfisherlevine, line 1: At a glance this looks fine, and is somewhat self-validating if it works, so I'm going to assume this is 👍 rather than scrutinising too much.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -305,7 +307,10 @@ def run(self, inputStats, inputDims):
         outputStats = {}
         success = True
         for expStats, dimensions in zip(inputStats, inputDims):
-            expId = dimensions['exposure']
+            expId = dimensions.get('exposure', dimensions.get('visit', None))
```
mfisherlevine, line 310: Should this be `visit` or `exposure`? Maybe they're the same here (and always will be?) but I'd have though it should be `exposure` at this point, no?

python/lsst/cp/verify/mergeResults.py
```diff
@@ -359,3 +364,76 @@ def verify(self, statisticsDictionary):
 
         """
         raise NotImplementedError("Subclasses must implement verification criteria.")
+
+
+class CpVerifyVisitExpMergeConnections(pipeBase.PipelineTaskConnections,
+                                       dimensions={"instrument", "visit"},
```
mfisherlevine, line 370: OK, I see it's visit basically everywhere. I hate this stuff, and am still not sure why that's the case, but I guess it really is a visit thing... One day it'll probably click for me, I guess.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -359,3 +364,76 @@ def verify(self, statisticsDictionary):
...
+        doc="Input camera.",
+        dimensions=["instrument", ],
+        isCalibration=True,
+    )
+
+    outputStats = cT.Output(
+        name="exposureStats",
+        doc="Output statistics.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "visit"],
```
mfisherlevine, line 391: Aren't the outputs still per-detector? Or do all the detectors get packed into the same output dict, hence the loss of that dimension?

examples/cpVerifyBfk.ipynb
mfisherlevine: One comment is that in cell 6 (I think) it crashes, and I think you've actually already fixed that error by updating the utils, so to have that render nicely you just need to rerun the notebook and push.

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+        name="detectorStats",
+        doc="Output statistics from cp_verify.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "visit", "detector"],
+    )
+
+
+class CpVerifyBfkConfig(CpVerifyStatsConfig,
+                        pipelineConnections=CpVerifyBfkConnections):
+    """Inherits from base CpVerifyStatsConfig.
```
mfisherlevine, line 72: I feel like the explicit inheritance in the definition above says this itself.

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+        super().setDefaults()
+        self.catalogStatKeywords = {'BRIGHT_SLOPE': None,
+                                    'NUM_MATCHES': None,
+                                   }  # noqa F841
+
+
+def exponentialModel(x, A, alpha, x0):
+    """An exponential model.
+    """
+    return A * np.exp(alpha * (x - x0))
```
mfisherlevine, line 91: Exponential fits without a constant can be pretty fickle, but as long as it's working well enough then it's probably fine.

examples/cpVerifyBfk.ipynb
mfisherlevine: I didn't find the naming of "numpy simple" for one of the fits very intuitive, and had to look at the code to see it was a polynomial fit. Fancy changing that to "poly fit" or something like that (and even including the order perhaps?)

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+            dictionaries of the statistics measured and their values.
+        """
+        outputStatistics = {}
+
+        matches = afwTable.matchXy(catalog, uncorrectedCatalog, self.config.matchRadiusPix)
+        outputStatistics['NUM_MATCHES'] = len(matches)
+
+        magnitude = []
+        sizeDiff = []
+        for source, uncorrectedSource, d in matches:
```
mfisherlevine, line 140: I wonder if this needs an `if not matches:` above it to deal with if that has failed? Or maybe that never happens.

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+            # This uses the simple difference in source moments.
+            sourceMagnitude = -2.5 * np.log10(source.getPsfInstFlux())
+            sourceSize = source['base_SdssShape_xx'] + source['base_SdssShape_yy']
+            uncorrectedSize = uncorrectedSource['base_SdssShape_xx'] + uncorrectedSource['base_SdssShape_yy']
```
mfisherlevine, line 144: Do we not need to add these in quadrature? (Or if they're already squared, or whatever, you know what I mean, do we need to sqrt after adding?)

Also, I wonder if there's value in reporting these separately as we can see size correction along the axes, given their independent, and there's physically motivated reasons for them to differ (especially if things are going wrong). Not saying not to do it combined as well, but separate could be useful too.

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+
+            magnitude.append(float(sourceMagnitude))
+            sizeDiff.append(float(uncorrectedSize - sourceSize))
+
+        mask = np.isfinite(magnitude) * np.isfinite(sizeDiff)
+        if 'BRIGHT_SLOPE' in self.config.catalogStatKeywords:
+            exponentialFit = least_squares(modelResidual, [0.0, 0.0, 0.0],
```
mfisherlevine: Exactly zero is not always the most robust starting value to use, especially if, for example, you know it's always going to be positive. Even just bumping to 0.1 might help for some cases.

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+                                           loss='cauchy')
+
+            outputStatistics['BRIGHT_SLOPE'] = float(exponentialFit.x[1])
+            outputStatistics['FIT_SUCCESS'] = exponentialFit.success
+            self.debugFit('brightSlope', magnitude, sizeDiff, exponentialFit.x)
+
+        outputStatistics['MAGNITUDES'] = magnitude
+        outputStatistics['SIZE_DIFF'] = sizeDiff
+
+        # This should have a debug view
```
mfisherlevine: Plan to add this?

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+
+            while True:
+                ans = input(prompt).lower()
+                if ans in ("", " ", "c",):
+                    break
+                elif ans in ("p", ):
+                    import pdb
+                    pdb.set_trace()
+                elif ans in ("h", ):
+                    print("[h]elp [c]ontinue [p]db e[x]itDebug")
```
mfisherlevine, line 202: I'm still sad we haven't centralised this (and added a `try: import ipdb as pdb; except import pdb` block in the process).

python/lsst/cp/verify/verifyBfk.py
```diff
@@ -0,0 +1,237 @@
...
+            A boolean indicating if all tests have passed.
+        """
+        verifyStats = {}
+        success = True
+        catalogVerify = {}
+        catStats = statisticsDict['CATALOG']
+
+        catalogVerify['BRIGHT_SLOPE'] = False
+        catalogVerify['NUM_MATCHES'] = False
+        # These values need justification.
```
mfisherlevine, line 233: People should be rewarded for leaving comments like this, rather than punished, but reading it I still can't really help but ask... so, what's the justification for these values? (sorry!)

examples/cpVerifyBfk.ipynb
czwa: Agree.  "polynomial (N=2)" is the new label.  90% of the time, I use the notebook to debug what I'm doing, then have a clean-up pass to make it presentable.  This one slipped through.

python/lsst/cp/verify/mergeResults.py
czwa: Unfortunately, BFK verify needs to run on catalog data, so it needs to be processed through `ImageCharacterization`.  That task converts `exposure` to `visit`, which in turn required a `Visit` set of merge tasks.  The code is functionally identical to the `exposure` version, so I added this to get the correct value from the dimension data.

python/lsst/cp/verify/mergeResults.py
czwa: As above.

python/lsst/cp/verify/mergeResults.py
czwa: The `Stats` task generates `{instrument, exposure (or visit), detector}` data.  The `ExpMerge` tasks merge all detectors to full exposure statistics, and then the `RunMerge` merges all exposures into a full evaluation.  It's the only way I could envision the hierarchical nature of the verify step.

python/lsst/cp/verify/verifyBfk.py
czwa: I'd rather have a silly docstring than someone (me) skipping the inheritance and wondering what's going on.

python/lsst/cp/verify/verifyBfk.py
czwa: The assumption is that at low flux levels, the size difference is zero.  Even if we don't get there (pretend spot data that only has bright sources), there should still be a theoretical flux that has zero difference.

python/lsst/cp/verify/verifyBfk.py
czwa: I've added such a block to be safe.

python/lsst/cp/verify/verifyBfk.py
czwa: I think they're already squared (units of pixel^2).  This also matches Craig and Alex Broughton's use (their `T` and `dT`).

python/lsst/cp/verify/verifyBfk.py
czwa: I've made the default `alpha0 = -0.01` to avoid any issues.  All of the derivatives are well behaved.

python/lsst/cp/verify/verifyBfk.py
czwa: Removed, because the `debugFit` method is what I was thinking about.

python/lsst/cp/verify/verifyBfk.py
czwa: `NUM_MATCHES > 10` exists because we need to be able to trust the fit.  In any real science image, this should be trivially met (the majority of my tests had ~200-300).
`BRIGHT_SLOPE < -0.5` is entirely arbitrary.  `alpha` must be negative (smaller size difference at larger magnitudes/fainter fluxes), but making it be simply negative ignores that the size change should be significant.  If `alpha =~ -0.01` we're essentially doing no correction.  Every other constraint I can think if is circular, because it requires trying to figure out how much smaller the kernel should make things.
I am open to any suggestion on this.

mfisherlevine: Yeah, I don't really have any great suggestions, just had to comment on the comment really. Maybe just pasting this into a comment and replacing the `# These values need justification.` line would do it. And that way, if there's verifications failing in future, we can see that perhaps these values should be adjusted, as opposed to tests failing. (The other potential problem I can imagine is that if there's not much BF present in the first place (due to low depth in the observations, for example), or different sensors, (because we're camera-agnostic, haha), then this could cause a fail here, so noting that these values shouldn't be taken as gospel would be good).

PR #9, opened 2021-11-15
DM-32470: Dark verification in OCPS calibration script from DM-31897 fails because NOISe is too low with respect nominal readnoise

czwa:
This change adds the possibility of pulling values from the task metadata.  This code path is then used to pull the measured scatter in the overscan as an estimate the read noise.  This should resolve the issue of the camera read noise not matching the quoted camera value.

### DIFF
--- a/python/lsst/cp/verify/verifyBias.py
+++ b/python/lsst/cp/verify/verifyBias.py
@@ -37,6 +37,7 @@ class CpVerifyBiasConfig(CpVerifyStatsConfig,
                                   'NOISE': 'STDEVCLIP', }
 
         self.crImageStatKeywords = {'CR_NOISE': 'STDEV', }  # noqa F841
+        self.metadataStatKeywords = {'RESIDUAL STDEV': 'AMP', }  # noqa F841
 
 
 class CpVerifyBiasTask(CpVerifyStatsTask):
@@ -68,18 +69,25 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
         """
         detector = exposure.getDetector()
         ampStats = statisticsDict['AMP']
+        metadataStats = statisticsDict['METADATA']
 
         verifyStats = {}
         success = True
         for ampName, stats in ampStats.items():
             verify = {}
-
+            readNoiseMatch = True
             # DMTN-101 Test 4.2: Mean is 0.0 within noise.
             verify['MEAN'] = bool(np.abs(stats['MEAN']) < stats['NOISE'])
 
             # DMTN-101 Test 4.3: Clipped mean matches readNoise.
-            amp = detector[ampName]
-            verify['NOISE'] = bool(np.abs(stats['NOISE'] - amp.getReadNoise())/amp.getReadNoise() <= 0.05)
+            readNoise = detector[ampName].getReadNoise()
+            if 'RESIDUAL STDEV' in metadataStats and ampName in metadataStats['RESIDUAL STDEV']:
+                overscanReadNoise = metadataStats['RESIDUAL STDEV'][ampName]
+                if ((overscanReadNoise - readNoise)/readNoise > 0.05):
+                    readNoiseMatch = False
+                readNoise = overscanReadNoise
+
+            verify['NOISE'] = bool(np.abs(stats['NOISE'] - readNoise)/readNoise <= 0.05)
 
             # DMTN-101 Test 4.4: CR rejection matches clipped mean.
             verify['CR_NOISE'] = bool(np.abs(stats['NOISE'] - stats['CR_NOISE'])/stats['CR_NOISE'] <= 0.05)
@@ -89,9 +97,10 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
                 verify['PROCESSING'] = False
 
             verify['SUCCESS'] = bool(np.all(list(verify.values())))
-
             if verify['SUCCESS'] is False:
                 success = False
+            # This isn't something to fail on yet.
+            verify['READ_NOISE_CONSISTENT'] = readNoiseMatch
 
             verifyStats[ampName] = verify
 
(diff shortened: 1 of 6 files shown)


python/lsst/cp/verify/verifyBias.py
```diff
@@ -37,6 +37,7 @@ def setDefaults(self):
                                   'NOISE': 'STDEVCLIP', }
 
         self.crImageStatKeywords = {'CR_NOISE': 'STDEV', }  # noqa F841
+        self.metadataStatKeywords = {'RESIDUAL STDEV': 'AMP', }  # noqa F841
```
plazas, line 39: Other keys, such as `CR_NOISE`, have `_` in between words. Would it be better to also add an underscore here to be consistent?

python/lsst/cp/verify/verifyBias.py
```diff
@@ -68,18 +69,25 @@ def verify(self, exposure, statisticsDict):
...
-
+            readNoiseMatch = True
             # DMTN-101 Test 4.2: Mean is 0.0 within noise.
             verify['MEAN'] = bool(np.abs(stats['MEAN']) < stats['NOISE'])
 
             # DMTN-101 Test 4.3: Clipped mean matches readNoise.
-            amp = detector[ampName]
-            verify['NOISE'] = bool(np.abs(stats['NOISE'] - amp.getReadNoise())/amp.getReadNoise() <= 0.05)
+            readNoise = detector[ampName].getReadNoise()
+            if 'RESIDUAL STDEV' in metadataStats and ampName in metadataStats['RESIDUAL STDEV']:
```
plazas, line 92: Should we add a comment here, or above, explaining `RESIDUAL STDEV` more? For example, saying that it is the standard deviation of the overscan region after the overscan has been removed., and that this should be identical to the true read noise (according to an explanation I got from you once). This given that the name doesn't seem to be that descriptive.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -68,18 +69,25 @@ def verify(self, exposure, statisticsDict):
...
+                overscanReadNoise = metadataStats['RESIDUAL STDEV'][ampName]
```
plazas, line 93: OK, here you assign the per amp `RESIDUAL STDEV`  to a variable with a more descriptive name, but maybe a comment would still be good.

python/lsst/cp/verify/verifyBias.py
```diff
@@ -89,9 +97,10 @@ def verify(self, exposure, statisticsDict):
                 verify['PROCESSING'] = False
 
             verify['SUCCESS'] = bool(np.all(list(verify.values())))
-
             if verify['SUCCESS'] is False:
                 success = False
+            # This isn't something to fail on yet.
```
plazas: So should this actually be a `TODO` with an associated ticket, or should we provide more information? I'm trying to think of our future selves when we come back to this part of the code and wonder about this (i.e., when/if should we fail on this?)

python/lsst/cp/verify/verifyDark.py
```diff
@@ -67,24 +68,39 @@ def verify(self, exposure, statisticsDict):
...
-
+            readNoiseMatch = True
             # DMTN-101 Test 5.2: Mean is 0.0:
             verify['MEAN'] = bool(np.abs(stats['MEAN']) < stats['NOISE'])
 
             # DMTN-101 Test 5.3: Clipped mean matches readNoise
-            amp = detector[ampName]
-            verify['NOISE'] = bool(np.abs(stats['NOISE'] - amp.getReadNoise())/amp.getReadNoise() <= 0.05)
+            readNoise = detector[ampName].getReadNoise()
+            if 'RESIDUAL STDEV' in metadataStats and ampName in metadataStats['RESIDUAL STDEV']:
```
plazas, line 92: Same comments in this "Dark" file as in the "Bias" one

python/lsst/cp/verify/verifyStats.py
```diff
@@ -176,14 +195,16 @@ def __init__(self, **kwargs):
...
-    def run(self, inputExp, camera):
+    def run(self, inputExp, camera, taskMetadata=None):
         """Calculate quality statistics and verify they meet the requirements
         for a calibration.
 
         Parameters
         ----------
         inputExp : `lsst.afw.image.Exposure`
             The ISR processed exposure to be measured.
+        taskMetadata : `lsst.daf.base.PropertySet`, optional
```
plazas, line 206: Is `RESIDUAL STDEV` the only statistic from the metadata that we currently use? Are there any other statistics in the metadata that could be useful?  Where would we need to look if we wanted to see all the contents of this metadata object to see if there's something else that might be useful?

plazas (approved):
Look good, thank you for working on this fix. Some clarifications requested

python/lsst/cp/verify/verifyBias.py
czwa: This comes from a choice in `ip_isr`.  I wanted to keep changes out of that.  I had thought about moving this from the task metadata to the exposure metadata, and then making the metadata stats code look there instead of the task metadata (removing that connection).  However, this wouldn't be available for previously processed data, and so there'd always be two versions.

This was a design decision, and I'm still not sure I made the right one.

python/lsst/cp/verify/verifyBias.py
czwa: Explanation added here and in verifyDark.  Minor differences also removed in that check (dark checked that `overscanReadNoise` was found, bias did not).

python/lsst/cp/verify/verifyBias.py
czwa: I've updated the comment.  I think we would only want to fail on this if we implement versioned cameras.

python/lsst/cp/verify/verifyStats.py
czwa: It's produced by `IsrTask`, and contains a variety of per-amp measurements at different stages of processing.  It also holds some statistics about the overscan.  It's stored there because when I reworked the task, I wanted to save some of the information that we were measuring that wasn't being saved anywhere.  This gets back to the design decision comment above; it's probably worth thinking about if this information should be written to the exposure metadata as well.

plazas: Maybe we can discuss this more in one of our meetings, especially if we decide that we want to use these statistics as stability metrics or something like that.  In the end, I'm just concerned about 1) are we saving potentially useful information in this metadata, 2) if so, how can we easily access it for plotting/use/further analysis?

PR #10, opened 2021-11-19
DM-31060: Use log.warning rather than deprecated log.warn

timj:


czwa (approved):
DM-32669 will fix the warnings.

PR #11, opened 2021-11-19
DM-29326: Fix logic of the doVignette option in isrTask

czwa:
I missed a vignette case here.

This also fixes a new build error related to scipy.


plazas (approved):
Looks good, sorry I missed it the first time too. I looked for more instances of `vignetteExposure` and it seems that there are not more.

PR #12, opened 2021-11-24
DM-32682: Use more general APIs for task metadata

timj:
This makes things more compatible with the upcoming TaskMetadata class.

One wrinkle here is that the pipeline requires task metadata so when we migrate away from PropertySet we will have to remember to change the pipeline storage class definition.

### DIFF
--- a/tests/test_verifyStats.py
+++ b/tests/test_verifyStats.py
@@ -27,8 +27,6 @@ import lsst.ip.isr.isrMock as isrMock
 import lsst.cp.verify as cpVerify
 import lsst.ip.isr.isrFunctions as isrFunctions
 
-from lsst.daf.base import PropertySet
-
 
 def updateMockExp(exposure, addCR=True):
     """Update an exposure with a mask and variance plane.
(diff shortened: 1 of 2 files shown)


tests/test_verifyStats.py
```diff
@@ -186,11 +184,11 @@ def setUp(self):
 
         # Use this to test the metadataStats code, as this is the case
         # it's designed to fix.
-        metadataContents = PropertySet()
-        metadataContents.setFloat("RESIDUAL STDEV C:0,0", 12.0)
-        metadataContents.setFloat("RESIDUAL STDEV", 24.0)
-        self.metadata = PropertySet()
-        self.metadata.setPropertySet("subGroup", metadataContents)
+        metadataContents = {}
```
ktlim: I guess this is the least-functional type of metadata that could be set, and hence the lowest common denominator, but when `TaskMetadata` is available, wouldn't it be better to use it instead to potentially provide a more complete and accurate test?

ktlim (approved):

tests/test_verifyStats.py
timj: Yes, but I didn't want to have the test code using the internal private state of pipe_base.

timj: Although of course I can just use `TaskMetadata` because when I merge pipe_base it will be available even if it's not used by `Task` by default.

timj:
@ktlim I've made it use `TaskMetadata` now.

PR #13, opened 2022-01-18
DM-33300: Drop connections that aren't needed in cp_verify

czwa:
Discard catalog inputs when no catalog input is needed.


plazas (approved):
This seems to have fixed the problem I was having with the OCPS calibration script failing the verification step (see the Jira comments). Thanks! I still need to run the OCPS script again when the camera is available to do so. 
What is the purpose of these catalogs? When should they be used (and how are they generated?), and why is it OK to just not use them when they are not available?

PR #15, opened 2022-04-21
DM-30170: Define crosstalk tests for cp_verify

czwa:
Add task to run validation against a calibration instead of against processed exposures, and subclass that for crosstalk.
Add unit test and example notebooks.


python/lsst/cp/verify/verifyCalib.py
```diff
@@ -0,0 +1,193 @@
...
+              TEST: boolean
+              TEST2: boolean
+            Amp2:
+            Amp3:
+        SUCCESS: boolean
+
+        """
+        outputStats = {}
+
+        outputStats['DETECTOR'] = self.detectorStatistics(inputCalib)
```
erykoff: I don't see how this could create a dict that resolves to the yaml described in the docstring.  Roots would be `DETECTOR`, `VERIFY`, `SUCCESS`, and not `AMP`, `DET`, `CATALOG`.

erykoff: Also below there's something about `N_VALID`?

erykoff (changes requested):
In my reading the code doesn't match the docstrings.  Am I missing something?

python/lsst/cp/verify/verifyCalib.py
czwa: This is a copy from the `CpVerifyStatsTask` block, and I forgot to clean it up to match reality.  I've removed the entries that do not exist, and renamed `DETECTOR` to `DET` to match how `verifyStats` names things.  `N_VALID` is one of the `STAT` entries, and I've updated the text to note that.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -439,3 +440,139 @@ class CpVerifyVisitRunMergeTask(CpVerifyRunMergeTask):
...
+        -----
+        The outputStats should have a yaml representation as follows.
+
+        VERIFY:
+          DetectorId1:
+            VERIFY_MEAN: boolean
+            VERIFY_SIGMA: boolean
+          DetectorId2:
+            [...]
+          MEAN_UNIMODAL: boolean
```
erykoff: Where do these UNIMODAL values come from?

czwa: That was me confusing myself by copying a docstring and incompletely editing it.  The new docstring now represents what is actually constructed, and use generic `TEST_VALUE` style names.

erykoff (approved):

PR #18, opened 2022-07-29
DM-30171: Define PTC tests for cp_verify

plazas:


python/lsst/cp/verify/verifyCalib.py
```diff
@@ -86,15 +96,33 @@ class CpVerifyCalibTask(pipeBase.PipelineTask, pipeBase.CmdLineTask):
     ConfigClass = CpVerifyCalibConfig
     _DefaultName = 'cpVerifyCalib'
 
-    def run(self, inputCalib):
+    def runQuantum(self, butlerQC, inputRefs, outputRefs):
```
czwa: This `runQuantum` doesn't do anything that the default wouldn't, so it doesn't need to be here.

python/lsst/cp/verify/verifyCrosstalk.py
```diff
@@ -73,7 +73,7 @@ def detectorStatistics(self, inputCalib):
 
         return outputStatistics
 
-    def verify(self, calib, statisticsDict):
+    def verify(self, calib, statisticsDict, **kwargs):
```
czwa: This should match the signature as it is in `verifyCalib.py`.

python/lsst/cp/verify/verifyPtc.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # Overwrite verifyDetStats with final boolean test over all amps
+        verifyDetStatsFinal = {}
+        for testName in verifyDetStats:
+            testBool = bool(np.all(list(verifyDetStats[testName])))
+            # Save the tests that failed
+            if not testBool:
+                verifyDetStatsFinal[testName] = bool(np.all(list(verifyDetStats[testName])))
+
+        return verifyDetStatsFinal, bool(success)
```
czwa, line 246: Github is complaining about the cases of the booleans in the VerifyDefect.yaml pipeline definition, but doesn't let me put a comment there, so it's here instead.

python/lsst/cp/verify/verifyPtc.py
```diff
@@ -0,0 +1,211 @@
...
+        ----------
+        inputCalib : `lsst.ip.isr.IsrCalib`
+            The calibration to verify.
+
+        Returns
+        -------
+        outputStatistics : `dict` [`str`, scalar]
+            A dictionary of the statistics measured and their values.
+        """
+        return {}
```
czwa, line 113: Can this be used to store the values that are being tested?  I know this duplicates information from the PTC dataset, but it will make it easier to migrate that information into a database in the future.

czwa (approved):
A few style notes, but I don't think there's any major issues.

python/lsst/cp/verify/verifyPtc.py
plazas: Yes. Is there an example of this in other "verify" tests? Should they be called something like "PTC_GAIN", "PTC_NOISE", etc. ?

plazas: I'm thinking about putting these in a new ``amplifierStatistics`` method since the relevant quantities are per amp (gain, turnoff, noise, a00).

PR #21, opened 2023-01-13
DM-37587: Construct flats for LATISS/January 12 data

czwa:
Update FLAT verify notebook.


examples/cpVerifyFlat.ipynb
```diff
@@ -19,145 +27,121 @@
...
     "\n",
     "# Which calibration type to analyse.\n",
     "calibType = 'flat'\n",
-    "physical_filter = 'empty~empty'\n",
+    "physical_filter = 'BG40_65mm_1~empty'\n",
     "# Camera\n",
     "cameraName = 'LATISS'\n",
     "\n",
     "# Set which display to use.\n",
     "afwDisplay.setDefaultBackend(\"astrowidgets\")\n",
```
plazas: Astrowidgets don't work at the USDF RSP, for example. Should this be in a try block with an alternative display in case of failure?

plazas (approved):
I think it looks good. I left a comment about a try block with Astrowidgets. in general, I think all these notebooks could be more pedagogical,  with cells with different markdown text to explain things. Like the DP0 and/or Stack Club tutorials. Especially if we intend for other people beyond DM (e.g., users like the observer specialists) to use them. So maybe we should file a ticket to do that documentation...

examples/cpVerifyFlat.ipynb
czwa: I've added a try, and asserts to the cells the use that display so they are skipped.  I've also added https://jira.lsstcorp.org/browse/DM-37592 to cover the documentation aspects you bring up.

PR #23, opened 2023-03-29
DM-36747: cp_verify noise checks cannot deal with measured noise smaller than the nominal read noise

czwa:
Remove abs calls in NOISE tests.


aferte:
Looks good to me.

PR #24, opened 2023-04-05
DM-38569: Move cp_verify notebooks out of examples/

czwa:
As part of this ticket, I've also updated the notebooks to use the updated style.  They have also been more fully parameterized, and should work with papermill.


plazas (approved):
The notebooks have been moved to a new location, as per the ticket requests. I have asked for clarification in the comments that how to use `papermill`, and the addressing of minor bugs, reported in Jira, before merging.

PR #31, opened 2023-10-05
DM-38563: Compare LATISS Defects generated from combined images

plazas:


python/lsst/cp/verify/verifyDefects.py
```diff
@@ -20,43 +20,190 @@
...
+        name="uncorrectedSrc",
+        doc="Input catalog without correction applied.",
+        storageClass="SourceCatalog",
+        dimensions=["instrument", "visit", "detector"],
+    )
+    camera = cT.PrerequisiteInput(
+        name="camera",
+        storageClass="Camera",
+        doc="Input camera.",
+        dimensions=[
```
czwa: I'm not to concerned about which formatting style we use, but it should be consistent.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -20,43 +20,190 @@
...
+        }
+        self.uncorrectedImageStatKeywords = {
+            "UNC_DEFECT_PIXELS": "NMASKED",  # noqa F821
+            "UNC_OUTLIERS": "NCLIPPED",
+            "UNC_MEDIAN": "MEDIAN",
+            "UNC_STDEV": "STDEVCLIP",
+            "UNC_MIN": "MIN",
+            "UNC_MAX": "MAX",
+        }
+        # initialize the dictionaries. They will get replaced.
```
czwa: Can you add something about "These config options need to have a key/value pair to run verification analysis, but the contents of that pair are not used"?  I'm sure I'll be confused next time I read this bit of code, and that should help.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -20,43 +20,190 @@
...
 
 class CpVerifyDefectsTask(CpVerifyStatsTask):
     """Defects verification sub-class, implementing the verify method.
 
     This also applies additional image processing statistics.
     """
+
     ConfigClass = CpVerifyDefectsConfig
-    _DefaultName = 'cpVerifyDefects'
+    _DefaultName = "cpVerifyDefects"
```
czwa, line 123: Thank you for clearing up my inconsistent use of ' and ".  It's the same as the formatting issue, and I know I'm consistently inconsistent.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -117,25 +279,57 @@ def verify(self, exposure, statisticsDict):
...
+        # Detection tests from DM-38563, before and after defects.
+        verifyStatsCat["NUMBER_DETECTIONS"] = bool(
+            catStats["NUM_OBJECTS_BEFORE"] > catStats["NUM_OBJECTS_AFTER"]
+        )
+
+        verifyStatsCat["SUCCESS"] = bool(np.all(list(verifyStatsCat.values())))
+        if verifyStatsCat["SUCCESS"] is False:
+            successCat = False
+
+        success = successDet & successAmp & successCat
```
czwa, line 330: I'm not convinced `successAmp` needs to be included, but if it doesn't cause problems, there's no reason to remove it, I think.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -67,10 +75,11 @@ class CpVerifyStatsConnections(pipeBase.PipelineTaskConnections,
         name="camera",
         storageClass="Camera",
         doc="Input camera.",
-        dimensions=["instrument", ],
+        dimensions=[
```
czwa: Same as above.

czwa (approved):

python/lsst/cp/verify/verifyDefects.py
plazas: Agree; this was probably a consequence of running `black`.

PR #33, opened 2023-11-07
DM-41633:  Generate empty~SDSSi_65mm flat for LATISS

czwa:
Fix detectorStatistics for verifyFlat.

Unify the signature and docstring for that method everywhere it is defined for VerifyStatsTask classes.


plazas (approved):
LGTM!

PR #34, opened 2024-01-12
DM-42160:  eo_pipe/cp_verify parity: dark

czwa:
Pass isrStatistics to output dictionary.

### DIFF
--- a/pipelines/_ingredients/VerifyDark.yaml
+++ b/pipelines/_ingredients/VerifyDark.yaml
@@ -11,12 +11,15 @@ tasks:
       doFlat: false
       doApplyGains: false
       doFringe: false
+      doCalculateStatistics: true
+      isrStats.doCopyCalibDistributionStatistics: true
   verifyDarkChip:
     class: lsst.cp.verify.CpVerifyDarkTask
     config:
       connections.inputExp: 'verifyDarkProc'
       connections.taskMetadata: 'verifyDarkApply_metadata'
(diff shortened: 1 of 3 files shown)


pipelines/_ingredients/VerifyDark.yaml
```diff
@@ -11,12 +11,15 @@ tasks:
       doFlat: false
       doApplyGains: false
       doFringe: false
+      doCalculateStatistics: true
+      isrStats.doCalibDistributionStatistics: true
```
arunkannawadi: Depending on whether or not you change the field name in `ip_isr`, this would change.

python/lsst/cp/verify/verifyStats.py
```diff
@@ -330,6 +346,9 @@ def run(
         else:
             outputStats["DET"] = {}
 
+        if self.config.useIsrStatistics:
+            outputStats["EXTRA"] = isrStatistics
```
arunkannawadi: Why is the key called `EXTRA` instead of something more specific?

arunkannawadi (approved):

python/lsst/cp/verify/verifyStats.py
czwa: Lack of imagination and a healthy fear of #dm-naming-things.  I've switched this to `ISR`, as these are the quantities measured during that stage.

PR #35, opened 2024-01-17
DM-42238: Add a check of consistency whether the noise measurements from the PTC fit and the overscan

Alex-Broughton:
…verscan

### DIFF
--- a/python/lsst/cp/verify/verifyPtc.py
+++ b/python/lsst/cp/verify/verifyPtc.py
@@ -240,6 +240,17 @@ class CpVerifyPtcTask(CpVerifyCalibTask):
             verify['PTC_GAIN'] = bool(diffGain < self.config.gainThreshold)
             verify['PTC_NOISE'] = bool(diffNoise < self.config.noiseThreshold)
 
+            # Check that the noises measured in cpPtcExtract do not evolve
+            # as a function of flux.
+            # We check that the reduced chi squared statistic between the
+            # noises and the mean of the noises less than 1.25 sigmas
+            mask = calib.expIdMask[ampName]
+            noiseList = calib.noiseList[ampName][mask]
+            expectedNoiseList = np.zeros_like(noiseList) + np.mean(noiseList)
+            chiSquared = np.sum((noiseList - expectedNoiseList)**2 / np.std(noiseList))
+            reducedChiSquared = chiSquared / len(noiseList)
+            verify['NOISE_SIGNAL_INDEPENDENCE'] = bool(reducedChiSquared < 1.25)
+
             # DMTN-101: 16.3
             # Check that the measured PTC turnoff is at least greater than the
             # full-well requirement of 90k e-.


python/lsst/cp/verify/verifyPtc.py
```diff
@@ -194,6 +196,15 @@ def verify(self, calib, statisticsDict, camera=None):
             verify['GAIN'] = bool(diffGain < self.config.gainThreshold)
             verify['NOISE'] = bool(diffNoise < self.config.noiseThreshold)
 
+            # Check that the noises measured in cpPtcExtract do not evolve
+            # as a function of flux by measuring their correlation with
+            # signal level.
+            signals = calib.rawMeans[ampName]
+            noiseList = calib.noiseList[ampName]
+            params = np.polyfit(signals, noiseList, deg=1)
+            maxNoiseDiff = (signals[-1] - signals[0])*params[0]
```
youtsumi: This could be just an issue in naming convention but  maxNoiseDiff sounds `max( |noise - model| )`. But what you are doing here is `signal_{n-1}-signal_{0}`, which is not the maximum of difference between the noise measurement and the model. Is it your intention?

python/lsst/cp/verify/verifyPtc.py
```diff
@@ -194,6 +196,15 @@ def verify(self, calib, statisticsDict, camera=None):
...
+            verify['NOISE_SIGNAL_COV'] = bool(maxNoiseDiff < np.std(noiseList))
```
youtsumi: Also, if you took true max(|NoiseDiff|), I don't expect this test pass reliably since you are comparing it with 1 sigma. I think we need some other test. 

For example, difference between upper and lower 90% percentile of noises should be less than 3*sigma?

python/lsst/cp/verify/verifyPtc.py
```diff
@@ -229,7 +240,8 @@ def verify(self, calib, statisticsDict, camera=None):
             verifyStats[ampName] = verify
 
         # Loop over amps to make a detector summary.
-        verifyDetStats = {'GAIN': [], 'NOISE': [], 'PTC_TURNOFF': [], 'BFE_A00': []}
+        verifyDetStats = {'GAIN': [], 'NOISE': [], 'NOISE_SIGNAL_COV': [],
```
youtsumi: I am not sure if `NOISE_SIGNAL_COC` is the right word to describe the definition.  `NOISE_TEST`?

youtsumi (commented):
I have made a few comments.

python/lsst/cp/verify/verifyPtc.py
czwa: This is correct as written.  It's measuring the amount of noise change attributable to flux changes over the set of input fluxes.  This should be small relative to the measured read noise, so I think the associated `verify` test is correct as well.

python/lsst/cp/verify/verifyPtc.py
czwa: I also don't like the name.  `NOISE_SIGNAL_INDEPENDENCE` was my thought, as that's what this test should be checking.

czwa (approved):
The github comment at the bottom can be resolved by removing the "---" line it complains about.  This may have been done by another ticket before you get to it, though, so it may disappear after rebasing.

python/lsst/cp/verify/verifyPtc.py
Alex-Broughton: The point I am trying test here is that the slope of the noise is essentially zero w.r.t. the signal level.

Here I am allowing the slope to vary such that it the maximum deviation in noise from 0 signal to max signal is indistinguishable form simply "noise in the noise"

python/lsst/cp/verify/verifyPtc.py
youtsumi: This is true if the noise evolution can be approximated by a linear function. But I presume the deviation from the nominal noise starts to evolve in the middle of the flux. 
I'd rather compare the maximum deviation with the n times `mad_std( noiseList )` or compare the number of noises that are deviated from `mad_std` and an expected number from the Gaussian distribution. 
This might not work for some amplifiers that have multimodal behavior in noises though... But we could relax the limit if that's the case.

python/lsst/cp/verify/verifyPtc.py
youtsumi: Ah ok, I think I was thinking differently. That sounds reasonable to me.

python/lsst/cp/verify/verifyPtc.py
youtsumi: see the comment below.

python/lsst/cp/verify/verifyPtc.py
plazas: You could make a short comment in the code explaining the intent of the test.

czwa (approved):

PR #38, opened 2024-01-25
DM-42159:  eo_pipe/cp_verify parity: bias

czwa:
This ticket adds:
 * Code to calculate readout corner mean values.
 * Updated handlers for PTC data.

### DIFF
--- a/python/lsst/cp/verify/verifyBias.py
+++ b/python/lsst/cp/verify/verifyBias.py
@@ -58,7 +58,6 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
         # Docstring inherited
         outputStatistics = super().imageStatistics(exposure, uncorrectedExposure, statControl)
 
-        # These should be a config item, probably.
         boxSize = self.config.ampCornerBoxSize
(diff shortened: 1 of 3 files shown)


python/lsst/cp/verify/verifyBias.py
```diff
@@ -45,6 +54,32 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
     ConfigClass = CpVerifyBiasConfig
     _DefaultName = 'cpVerifyBias'
 
+    def imageStatistics(self, exposure, uncorrectedExposure, statControl):
+        # Docstring inherited
+        outputStatistics = super().imageStatistics(exposure, uncorrectedExposure, statControl)
+
+        # These should be a config item, probably.
```
erykoff: You mean the statistic as well as the box size?  Either fix this or remove the comment.

erykoff (approved):

python/lsst/cp/verify/verifyBias.py
czwa: Just the box size.  I've removed the comment as it needed to be made a config to pass the unit tests (the default box size is too large for the mocked image).

PR #40, opened 2024-01-29
DM-42165: eo_pipe/cp_verify parity: bfk

plazas:


czwa (dismissed):
I hit approve before rereading the ticket:  this also needs to add the x-corr/y-corr slopes, doesn't it?  Those should be simple linear fits to `{flux, cov10}` and `{flux, cov01}`.

czwa (changes requested):
Covariance slopes are missing.

plazas:
> I hit approve before rereading the ticket: this also needs to add the x-corr/y-corr slopes, doesn't it? Those should be simple linear fits to `{flux, cov10}` and `{flux, cov01}`.

Sorry for missing those, I thought we wanted to calculate them in analysis tools. The most recent commit adds these parameters.

czwa (approved):

PR #42, opened 2024-01-29
DM-42702: Add conversions and padding for bias repackDetStats serialization.

erykoff:


czwa (approved):
I think I have the dimension fix on a ticket as well, but this will certainly merge first.

czwa (approved):
Looks fine to me.

PR #43, opened 2024-02-21
DM-42732: Add missing bias shift data

czwa:
This also adds the updated amp correlation results.


enourbakhsh (commented):
Looks good. Would it be useful to unit test these changes?

enourbakhsh (approved):

czwa:
> Looks good. Would it be useful to unit test these changes?

I've linked this ticket to DM-42661, which will fix a number of similar issues that were brought up by the sprint.  That ticket will update the target result files that are produced by `ci_cpp` and re-enable the test that compares those targets to the files that are generated by the pipelines.

PR #46, opened 2024-04-12
DM-42927: Update cp_verify connections/classes/outputs for analysis_tools

czwa:
This reworks the output data products to use ArrowAstropy tables instead of yaml, but does not fully remove the existing yaml (which I've added DM-43893 to handle).


python/lsst/cp/verify/mergeResults.py
```diff
@@ -84,101 +145,167 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
         outputs = self.run(**inputs)
         butlerQC.put(outputs, outputRefs)
 
-    def run(self, inputStats, camera, inputDims):
+    def run(self, inputStats, inputDims, camera=None, inputResults=None, inputMatrix=None,):
         """Merge statistics.
 
         Parameters
         ----------
```
erykoff, line 151: If I'm reading the diff correctly you removed the Parameters instead of fixing them?

python/lsst/cp/verify/mergeResults.py
```diff
@@ -54,19 +73,61 @@ class CpVerifyExpMergeConnections(pipeBase.PipelineTaskConnections,
...
+        dtype=bool,
+        doc="Are there results tables to merge?",
+        default=False,
+    )
+
+    thisDimension = pexConfig.Field(
+        dtype=str,
+        doc="Dimension name for this input.",
+        default="detector",
+    )
```
erykoff, line 124: So this works because the quantum dimensions (default are `"instrument", "exposure"`) must not include whatever is listed here?  If this is the case, there should be a check in verify that `thisDimension` is not listed above.  
At the same time, I don't understand the name (`thisDimension`) or the docstring (`"Dimension name for this input."`).

python/lsst/cp/verify/mergeResults.py
```diff
@@ -84,101 +145,167 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
+        testTable = inputResults[0]  # This has the default set of columns.
+        defaults = {key: -1 for key in testTable.columns}
+
+        # Identify vector columns:
+        for column in defaults.keys():
+            if len(testTable[column].shape) > 1:
+                defaults[column] = max(defaults[column],
+                                       *[table[column].shape[1] for table in inputResults])
+
+        # Pad vectors shorter than this:
```
erykoff: This padding stuff always makes me nervous.  I actually don't know what vectors are being padded here (more comments would be helpful) and why some would be shorter.  If it's something like a missing amp then padding may be wrong because we don't know which are the missing amps (e.g. padding was causing off-by-one errors in the ptc previously).

python/lsst/cp/verify/mergeResults.py
```diff
@@ -238,137 +365,233 @@ class CpVerifyRunMergeConnections(pipeBase.PipelineTaskConnections,
...
-        inputs = butlerQC.get(inputRefs)
+    pass
 
-        dimensions = [dict(exp.dataId.required) for exp in inputRefs.inputStats]
-        inputs['inputDims'] = dimensions
 
-        outputs = self.run(**inputs)
-        butlerQC.put(outputs, outputRefs)
+# End ExpMerge/RunMerge
+# Begin ExpMergeByFilter/RunMergeByFilter
```
erykoff, line 469: If we want these comments here, there should be a space between them.  Or a space, then the comments, then another space and the new class?

python/lsst/cp/verify/verifyCalib.py
```diff
@@ -236,3 +278,57 @@ def verify(self, inputCalib, statisticsDict, camera=None, exposure=None):
             subclass.
         """
         raise NotImplementedError("Subclasses must implement verification criteria.")
+
+    def repackStats(self, statisticsDict, dimensions):
+        """Repack information into flat tables.
+
+        This method should be redefined in subclasses.
```
erykoff: If this should be redefined, why isn't it an ABC with no code?

python/lsst/cp/verify/verifyCrosstalk.py
```diff
@@ -126,3 +130,73 @@ def verify(self, calib, statisticsDict, camera=None):
             success = False
 
         return verifyStats, success
+
+    def repackStats(self, statisticsDict, dimensions):
+        """Repack information into flat tables.
+
+        This method should be redefined in subclasses.
```
erykoff: If this is a subclass redefinition, it doesn't actually need a docstring.  You can add a comment `# docstring inherited` to make that clear.

erykoff (commented):
I have done a first pass with a bunch of comments.  But more importantly, I don't have a picture of how this is supposed to work, let alone how it works.  A discussion may be in order, plus more documentation.

python/lsst/cp/verify/mergeResults.py
czwa: The `thisDimension` field is intended to be the dimension that we're merging over.  This allows the same class and code to be able to merge over a set of inputs with individual dimensions `{instrument, exposure, detector}` over `detector` and then to merge the outputs of that step (which now have dimensions `{instrument, exposure}`) over `exposure`.

python/lsst/cp/verify/mergeResults.py
czwa: No, I forgot to update this docstring.  Should be fixed now.

python/lsst/cp/verify/mergeResults.py
czwa: I think this is doing the wrong thing in at least some cases, but I haven't been able to debug that yet (this ticket: https://rubinobs.atlassian.net/browse/DM-43877).  For more standard vectors (the serial and parallel profile, for example) this seems to be working correctly.

python/lsst/cp/verify/mergeResults.py
erykoff: If it's the dimension to merge over, may I suggest a name of `mergeDimension`?

python/lsst/cp/verify/mergeResults.py
czwa: I've made the comments stick either to the first or last line as appropriate, and put two spaces between, which seems to keep flake8 happy.  I added them after editing the wrong nearly-identical class multiple times.

python/lsst/cp/verify/mergeResults.py
erykoff: What is going into this table exactly?  And what are the rows that need padding?
If the standard vectors don't need padding we shouldn't have padding here at all.  I'd rather let something crash here with unexpected input than pass on garbage.

python/lsst/cp/verify/verifyCalib.py
czwa: Another dropped docstring.  This is a basic repacker for the simplest cases, and more complicated cases will need to replace it.  Docstring updated here and in `verifyStats.py` as well.

python/lsst/cp/verify/mergeResults.py
czwa: It's possible this isn't needed.  I've added a raise here, and am seeing what ci_cpp yields.  If that passes, then the padding is fully unnecessary, and I'll remove it.  Otherwise, I'll see what is causing the lengths to differ.

czwa: ci_cpp ran without complaint, so I think this means the padding can be removed.

erykoff: I would not be shocked to find something around here crashing when we run the full camera, but then we can/should directly fix whatever is giving inconsistent outputs.  So removing the padding is the right thing to do.

czwa:
I've added a short introduction to how cp_verify works to the documentation directory.  This builds, so I have not made any significant errors in the formatting of that document, but do not intend for this to be a final draft.  As cp_verify is not currently part of the documentation build, I've added a new ticket DM-43993 that will finalize and correct this draft, add the remaining missing documentation, and ensure that cp_verify appears on pipelines.lsst.io.  I'm happy to rewrite this draft for clarity and factual errors, but would prefer to keep formatting and fine details for that future ticket.

doc/lsst.cp.verify/using-cp-verify.rst
```diff
@@ -0,0 +1,81 @@
...
+
+* ``crImageStatKeywords``:  these options define the tests that will be run after cosmic ray rejection has been applied.
+
+* ``normImageStatKeywords``:  these options define the tests that will be run after normalizing the image by the exposure time.
+
+* ``detectorStatKeywords``:  as all previous image analysis is done on a per-amplifier basis, these options define tests to be run on the full detector.  As these will nearly always depend on the calibration under consideration, there is no defined method in the base class, so subclasses will need to implement it.
+
+* ``metadataStatKeywords``:  these options define the tests that will be run on the input task_metadata input.  These tests need to be implemented by subclasses.
+
+* ``catalogStatKeywords``:  these options define the tests that will be run on the input catalog data.  As before, these tests need to be implemented by subclasses.
```
erykoff: What sort of catalogs are these?  And where are they input from?

doc/lsst.cp.verify/using-cp-verify.rst
```diff
@@ -0,0 +1,81 @@
+.. py:currentmodule:: lsst.cp.verify
+
+#####################################
+Verifying calibrations with cp_verify
+#####################################
+
+`cp_verify` is designed to make the verification of calibrations against the criteria listed in `DMTN-101 <https://dmtn-101.lsst.io>`_ as simple as possible.  However, this has resulted in the tasks being rather abstract, with no clear connection between steps.  This document should help explain how the tests are run, how those results are stored, and how they are passed to `analysis_tools` for conversion into metrics and plots.
```
erykoff: I know you don't want a full review of formatting and everything, but the standard is that each sentence gets its own line in the rst doc files.  This is important to do before the first merge so that subsequent changes are easier to read in the github diff.

doc/lsst.cp.verify/using-cp-verify.rst
```diff
@@ -0,0 +1,81 @@
...
+* ``normImageStatKeywords``:  these options define the tests that will be run after normalizing the image by the exposure time.
+
+* ``detectorStatKeywords``:  as all previous image analysis is done on a per-amplifier basis, these options define tests to be run on the full detector.  As these will nearly always depend on the calibration under consideration, there is no defined method in the base class, so subclasses will need to implement it.
+
+* ``metadataStatKeywords``:  these options define the tests that will be run on the input task_metadata input.  These tests need to be implemented by subclasses.
+
+* ``catalogStatKeywords``:  these options define the tests that will be run on the input catalog data.  As before, these tests need to be implemented by subclasses.
+
+The final two configuration options that control the behavior of the verification task are the ``useIsrStatistics`` configuration option, which defines if there are results from the ``isrStatistics`` input that should be considered, and the ``hasMatrixCatalog`` option, which indicates if a matrix catalog output will be constructed.
+
```
erykoff, line 82: Even after reading the following paragraph, I'm still not sure what a "matrix" is.  I think a concrete example of what a matrix quantity is would help a lot.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -84,101 +144,165 @@ def runQuantum(self, butlerQC, inputRefs, outputRefs):
...
         return pipeBase.Struct(
             outputStats=outputStats,
+            outputResults=outputResults,
+            outputMatrix=outputMatrix,
         )
 
-    def exposureStatistics(self, statisticsDict):
+    @staticmethod
+    def mergeTable(inputResults, newStats=None):
+        """Merge input tables, padding columns as needed.
```
erykoff: I don't think we're doing padding any more.

python/lsst/cp/verify/verifyFlat.py
```diff
@@ -205,3 +264,46 @@ def verify(self, detectorStatistics, statisticsDictionary):
         success = bool(np.all(list(verifyStats.values())))
 
         return {'EXP': verifyStats}, bool(success)
+
+    def pack(self, statisticsDict, dimensions, outKey):
+        """Repack information into flat tables.
+
+        This method should be redefined in subclasses.
```
erykoff: Should or may?

ups/cp_verify.table
```diff
@@ -8,5 +8,6 @@ setupRequired(pex_exceptions)
 setupRequired(pipe_tasks)
 setupRequired(utils)
 setupRequired(pipe_base)
+setupRequired(analysis_tools)
```
erykoff, line 11: Is this an okay dependency direction?

doc/lsst.cp.verify/using-cp-verify.rst
czwa: These are catalogs from `CharacterizeImageTask`.  The main case for this is in the brighter-fatter verification, where we expect that objects will have smaller sizes with BF correction than without.  Text updated to indicate that a prior step in the pipeline should generate these, and a pointer to `CharacterizeImageTask`.

doc/lsst.cp.verify/using-cp-verify.rst
czwa: Done here so the actual documentation ticket will have the cleaner diffs.

doc/lsst.cp.verify/using-cp-verify.rst
czwa: Updated text in the `outputMatrix` description:

> An example use case for this kind of matrix catalog is the crosstalk coefficients, which would be represented with columns for source and target detector and amplifier, with the value column containing the coefficient at which the source pixels imprint onto the target pixels.

python/lsst/cp/verify/verifyFlat.py
czwa: Comment removed here, retained as "should" in the base class (newly added as this method was skipped), with the clarification that that's only if new stats have been calculated that need to be `pack`ed.

ups/cp_verify.table
czwa: I believe so.  `analysis_tools` depends on a subset of the `cp_verify` dependencies, in addition to `daf_butler`, `skymap`, and `geom`.  Those three only depend on even more fundamental packages, so I think this is safe from circular issues.

PR #48, opened 2024-06-13
DM-44259: Update pipeline names and code for RFC-1013

erykoff:


pipelines/_ingredients/verifyBfk.yaml
```diff
@@ -0,0 +1,65 @@
...
+      connections.sourceCat: "verifyUncorrectedBfkSrc"
+      connections.backgroundModel: "verifyUncorrectedBfkBackground"
+      connections.outputSchema: "verifyUncorrectedBfk_schema"
+      doApCorr: false
+      doDeblend: false
+      doMeasurePsf: false
+  verifyBfkChip:
+    class: lsst.cp.verify.CpVerifyBfkTask
+    config:
+      connections.inputExp: "icExp"
```
czwa: Should these (coming from the `verifyBfkChar` task) be named `verifyBfkIcExp` or something?

python/lsst/cp/verify/__init__.py
```diff
@@ -30,3 +30,6 @@
 from .verifyFlat import *
 from .verifyCrosstalk import *
 from .verifyPtc import *
+from .verifyLinearity import *
```
czwa: Alphabetize.

czwa (approved):
I think this needs an update to the README.md in the pipelines directory to note that we're now testing pipelines, so anything new should be added to that test.

PR #49, opened 2024-06-17
DM-44548  Patches for calibration rehearsal 1 (CR1)

czwa:
Patches that were needed for the calibration rehearsal, that we now want to include on main:

- Adds LSSTComCamSim specific pipeline definitions.
- Adds LSSTComCamSim to the pipeline unit tests.
- Adds the ability to drop the `outputStats` connection to avoid butler dimension mismatches on this to-be-deprecated dataset.

### DIFF
--- a/python/lsst/cp/verify/mergeResults.py
+++ b/python/lsst/cp/verify/mergeResults.py
@@ -450,6 +450,10 @@ class CpVerifyRunMergeConnections(pipeBase.PipelineTaskConnections,
         if not self.config.hasMatrixCatalog:
             self.inputs.remove("inputMatrix")
             self.outputs.remove("outputMatrix")
+        if not self.config.hasInputResults:
+            self.inputs.remove("inputResults")
+        if self.config.doDropStats:
+            self.outputs.remove("outputStats")
 
 
 class CpVerifyRunMergeConfig(CpVerifyExpMergeConfig,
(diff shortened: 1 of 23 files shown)


python/lsst/cp/verify/mergeResults.py
```diff
@@ -116,6 +118,11 @@ class CpVerifyExpMergeConfig(pipeBase.PipelineTaskConfig,
         doc="Are there results tables to merge?",
         default=False,
     )
+    doDropStats = pexConfig.Field(
+        dtype=bool,
+        doc="Drop deprecated stats files?",
+        default=False,
```
erykoff, line 124: If these are deprecated, why is the default not True?  Also, is there a ticket to remove any deprecated files (if so it should be tagged here with a TODO; if not it should be filed and then tagged here with a TODO.)

erykoff (commented):
I'm not sure what the `doDropStats` is about; it doesn't seem to be used or tested that I can see.

python/lsst/cp/verify/mergeResults.py
czwa: They're not formally deprecated, and this was a quick-fix to avoid the problem.  
The problem:  When I updated `cp_verify` to get the `analysis_tools` development unblocked, I was more explicit about which tasks have filter dimensions and which don't.  This changed the `flatStats` dimensions to include `physical_filter`, matching the `flatResults` dataset, but causing conflicts with the existing definition in `/repo/embargo`.  Since the `flatStats` never gest used outside of `cp_verify`, but the `flatResults` do (they're the input to `analysis_tools`), having this dataset get dropped avoided the butler problem.
The correct path forward is to RFC the `*Stats` datasets to be deprecated and removed, and I don't think that will have any objections, but having this shortcut is helpful until I have the time for the proper RFC deprecation.

erykoff: Okay, then I would suggest updating the doc string here.  Also, I'm not sure where this is even set to `True`!

czwa: I think the pipeline that had this set to true was rebased out of existence due to RFC-1013.  I'll clarify the doc string.

PR #53, opened 2024-07-19
DM-45300: Begin deprecation of task metadata in cp_pipe/cp_verify

czwa:
Refactor the way we handle metadata statistics, in anticipation that we can get everything from the exposure headers.


erykoff (approved):
My one question is if `READ_NOISE` is the correct name.  Is this always from the overscan region?  If so, then `READ_NOISE` should probably be something like `OVERSCAN_READ_NOISE`.  But if it's from the whole bias frame in that case then ... maybe `READ_NOISE` is okay?

PR #54, opened 2024-08-15
DM-45464: Fix handling of deprecated taskMetadata

czwa:
Begin deprecation of taskMetadata in verifyStats.

### DIFF
--- a/tests/test_verifyStats.py
+++ b/tests/test_verifyStats.py
@@ -27,8 +27,6 @@ import lsst.ip.isr.isrMock as isrMock
 import lsst.cp.verify as cpVerify
 import lsst.ip.isr.isrFunctions as isrFunctions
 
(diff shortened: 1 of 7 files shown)


tests/test_verifyStats.py
```diff
@@ -211,12 +209,10 @@ def setUp(self):
                            'detector': detector.getName(),
                            }
 
-        # Use this to test accessing info from the TaskMetadata:
-        metadataContents = TaskMetadata()
+        # Populate the READ_NOISE into the exposure header
+        md = self.inputExp.getMetadata()
```
erykoff, line 213: If you want to be fancy and modern this can be `self.inputExp.metadata`.  But this is also fine.

erykoff (commented):
All the metadata overrides in the pipelines need to be removed as well.

erykoff (approved):

PR #55, opened 2024-08-23
DM-45664:  Fix incorrect definition for mosaics

czwa:
And flatVerify task products.


erykoff (approved):
So the X isn't needed anymore?  Good!

PR #57, opened 2024-09-19
DM-46356: Create new verification pipelines for IsrTaskLSST for LATISS

erykoff:


pipelines/_ingredients/verifyBiasLSST.yaml
```diff
@@ -0,0 +1,99 @@
+description: cp_verify BIAS calibration verification (LSST style).
```
czwa: This ticket merges after DM-45856, correct?  Are these on this ticket as well so they could have parallel development?

czwa: Same ordering comment as on DM-45856.

tests/test_pipelines.py
```diff
@@ -121,7 +122,13 @@ def test_cameras(self):
...
     @unittest.skipIf(not has_obs_lsst, reason="Cannot test LATISS pipelines without obs_lsst")
     def test_latiss_pipelines(self):
-        for pipeline in self._get_pipelines():
+        for pipeline in self._get_pipelines(exclude=[
+                # The following tasks are not part of the new pipelines.
+                "verifyDefectsIndividual.yaml",
+                # The following tasks will be added in the future.
+                "verifyCrosstalk.yaml",
+                "verifyBfk.yaml",
+        ]):
```
czwa, line 141: If I'm understanding this correctly, we will not be testing the legacy pipelines anymore.  Should we keep testing them for now, and deprecate that in the future when we've finished the transition?

czwa (approved):
I think most of this was covered on the other ticket.  One question on testing legacy pipelines for the transition period.

pipelines/_ingredients/verifyBiasLSST.yaml
erykoff: Yup, we can only test the LATISS pipelines on top of the _ingredients in DM-45856.

tests/test_pipelines.py
erykoff: The verifyBfk pipeline will be coming on this ticket.  I don't know if you're referring to the verification pipeline (which should be easy to adapt for crosstalk, I hope, even if it's verifying existing crosstalk!) or the construction pipeline (which is already in rough shape).

PR #58, opened 2024-09-20
DM-44545:  Design initial calibration report format

czwa:
This adds a command line `cpv_report.py` that generates a web accessible calibration report from a butler repo and list of collections into an output directory.


python/lsst/cp/verify/configs/report.yaml
```diff
@@ -0,0 +1,81 @@
...
+    cpPtcCore_ptcTurnoffPerAmp_FocalPlaneGeometryPlot:
+      description: "Measured PTC turnoff values."
+    cpPtcCore_ptcRowMeanVarianceSlopePerAmp_FocalPlaneGeometryPlot:
+      description: "TBA."
+    cpPtcDetCore_ptcPlot_GridPlot:
+      description: "TBA."
+  linearity:
+  crosstalk:
+  bfk:
+  cti:
```
sr525, line 81: Eventually we'll need to merge this and the version on DM-45843 if we want to move to one day using the same report template code.

python/lsst/cp/verify/report.py
```diff
@@ -0,0 +1,780 @@
...
+        norm = colors.BoundaryNorm(quantiles, cmap.N)
+
+        im = ax.imshow(array, norm=norm, interpolation='None', cmap=cmap, origin='lower')
+
+        divider = make_axes_locatable(ax)
+        cax = divider.append_axes("right", size="5%", pad=0.05)
+        fig.suptitle(title)
+        fig.colorbar(im, cax=cax)
+        fig.tight_layout()
+        fig.savefig(target_uri.path)
```
sr525, line 347: Do you want a plt.close here?

python/lsst/cp/verify/configs/report.yaml
czwa: That's my plan as well.  I think this is compatible with the DM-45843 version, with this having only the minimum necessary fields (just the description).

python/lsst/cp/verify/report.py
czwa: I don't know.  If I'm reading the changelog for `make_figure` correctly (https://github.com/lsst/utils/blob/7d0def3d61c752f736c8d2cc33852495fcc90713/doc/changes/DM-44725.md?plain=1), then I think no, because this figure is independent of `pyplot` in some fashion.  I've not added one, and hope that when the method ends, the `fig` drops out of scope and is cleared.

PR #59, opened 2024-10-02
DM-46357: Add verification pipelines for new IsrTaskLSST for LSSTComCam and LSSTComCamSim

erykoff:


czwa (approved):
I confused myself a few times, but I think I sorted out what I had wrong (ComCam and ComCamSim shouldn't look as similar as they do in my head).

PR #61, opened 2024-11-15
DM-42201: Add flat core analysis tools pipeline

Alex-Broughton:


czwa (approved):

python/lsst/cp/verify/mergeResults.py
```diff
@@ -477,6 +479,172 @@ class CpVerifyRunMergeTask(CpVerifyExpMergeTask):
...
+        doc="Input camera.",
+        dimensions=["instrument", ],
+        isCalibration=True,
+    )
+
+    outputStats = cT.Output(
+        name="exposureStats",
+        doc="Output statistics.",
+        storageClass="StructuredDataDict",
+        dimensions=["instrument", "exposure"],
```
czwa: This, and the other outputs, should have dimensions `["instrument", "detector"]`.

python/lsst/cp/verify/mergeResults.py
```diff
@@ -477,6 +479,172 @@ class CpVerifyRunMergeTask(CpVerifyExpMergeTask):
...
+    )
+    camera = cT.PrerequisiteInput(
+        name="camera",
+        storageClass="Camera",
+        doc="Input camera.",
+        dimensions=["instrument", ],
+        isCalibration=True,
+    )
+
+    outputStats = cT.Output(
```
czwa, line 596: This set look fine, compared to those above.

czwa (approved):

czwa:
This slipped my mind, but to add to this:  If you add these new data products to `python/lsst/cp/verify/configs/report.yaml`, they will automatically be added to the cpv reports.  The dataset_type_name is the key to the middle level dict, and should be (based on the analysis_tools changes) look like `cpFlatDetMergeCore_flatTestsByDate_CalibAmpScatterTool` using the `outputName`, the "test name" set under the `atools` header, and the output type.

czwa:
Sorry, `cpFlatDetMergeCore_flatTestsByDate_GridPlot`, I think is the right thing.

czwa:
This needs rebasing, which will hopefully get the `gitignore` update that will cause `python/lsst_cp_verify.dist-info/METADATA` to be ignored.  If not, it should be removed from git manually.  Otherwise, I'm happy here.

timj:
Rebasing won't cause the .dist-info to disappear -- the git ignore file does nothing after the fact. The commit that added it will have to be cleaned up.

Alex-Broughton:
Will do; will push my updated local branch when I finish fixing another ci_cpp error. Thanks for pointing it out

PR #63, opened 2025-01-09
DM-48380: Identify why summit OCPS is having issues with LATISS calibration checkout

czwa:
Disable brighter-fatter for LATISS verifyFlat.


erykoff:
This is fine such as it is, but this is set to false in `isrLSST.py` for `LATISS`.  ooooh, but the pipeline yaml takes precedence doesn't it.  Sigh.  https://github.com/lsst/cp_verify/blob/main/pipelines/_ingredients/verifyFlatLSST.yaml#L22

erykoff (approved):

PR #67, opened 2025-02-23
DM-48955: Add verification pipelines and bps templates for LSSTComCam illumination corrections.

erykoff:


pipelines/_ingredients/verifyIlluminationCorrectionLSST.yaml
```diff
@@ -0,0 +1,117 @@
...
+    class: lsst.fgcmcal.fgcmFitCycle.FgcmFitCycleTask
+    config:
+      doMultipleCycles: true
+      multipleCyclesFinalCycleNumber: 5
+      superStarForceZeroMean: true
+  verifyIlluminationCorrectionFgcmOutputIlluminationCorrection:
+    class: lsst.fgcmcal.fgcmOutputIlluminationCorrection.FgcmOutputIlluminationCorrectionTask
+    config:
+      connections.cycleNumber: 5
+      connections.illumination_corrections: "verifyIlluminationCorrection"
```
czwa, line 81: If I'm following through correctly, this should be a zero (or 1.0) image, right?  If we're not adding a `verifyIlluminationCorrection.py` python class today, can we at least ticket that and link it to this ticket?

czwa (approved):
One request to ticket a future verification step for these products.

PR #70, opened 2025-03-31
DM-49730: Table merge error in verifyLinearizer and verifyPtc and verifyBiasExpMerge

czwa:
This change ensures that the columns in the merged astropy tables have the maximum size of all inputs.  Entries that are smaller than this size are padded with `np.nan` up to that size.
In order to add a unit test for this functionality, I had to update the unit test to run the table merges.  This resulted in some additional changes to ensure that the failure cases expected are still generated.

### DIFF
--- a/python/lsst/cp/verify/mergeResults.py
+++ b/python/lsst/cp/verify/mergeResults.py
@@ -303,12 +303,13 @@ class CpVerifyExpMergeTask(pipeBase.PipelineTask):
             try:
                 outputResults = vstack(inputResults)
             except TableMergeError:
+                # astropy requires everything in a column to have the
+                # same shape.  Enforce that here.
                 import numpy as np
-                # Ugh.
                 max_shapes = {}
                 columns_to_fix = set()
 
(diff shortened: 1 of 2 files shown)


python/lsst/cp/verify/mergeResults.py
```diff
@@ -300,7 +300,61 @@ def mergeTable(inputResults, newStats=None):
...
+                # astropy requires everything in a column to have the
+                # same shape.  Enforce that here.
+                import numpy as np
+                max_shapes = {}
+                columns_to_fix = set()
+
+                # Scan the data we're merging to find the max sizes.
+                for result in inputResults:
+                    for column in result.columns:
+                        size = result[column].shape
```
Alex-Broughton: Maybe its just me, but `size` ~ `len()`, so maybe this variable should be `shape` instead

Alex-Broughton:
One small comment, but otherwise looks good to me

Alex-Broughton (approved):

PR #71, opened 2025-04-03
DM-49219: Fix cp_verify noise tests

czwa:
This switches to use the calculated header values.

### DIFF
--- a/python/lsst/cp/verify/verifyBias.py
+++ b/python/lsst/cp/verify/verifyBias.py
@@ -24,6 +24,7 @@ import lsst.afw.math as afwMath
 
 from lsst.geom import Point2I, Extent2I, Box2I
 from lsst.pex.config import Field
+from lsst.ip.isr.isrFunctions import getExposureReadNoises, getExposureGains
 from .verifyStats import CpVerifyStatsConfig, CpVerifyStatsTask, CpVerifyStatsConnections
 
 __all__ = ['CpVerifyBiasConfig', 'CpVerifyBiasTask']
@@ -108,6 +109,8 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
 
         verifyStats = {}
         success = True
+        gains = getExposureGains(exposure)
+        readNoises = getExposureReadNoises(exposure)
         for ampName, stats in ampStats.items():
             verify = {}
 
@@ -128,7 +131,7 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
             # measured on the bulk of the image as the overscan
             # correction should be an optimal fit to the overscan
             # region, but not necessarily for the image region.
-            readNoise = exposure.getMetadata().get(f"LSST ISR READNOISE {ampName}", np.nan)
+            readNoise = readNoises[ampName]
             verify['NOISE'] = bool((stats['NOISE'] - readNoise)/readNoise <= 0.05)
 
             # DMTN-101 Test 4.4: CR rejection matches clipped mean.
@@ -152,7 +155,7 @@ class CpVerifyBiasTask(CpVerifyStatsTask):
             # if the metadataStats contain the required entry.  This
             # is in ADU (the serial overscan is measured prior to gain
             # normalization), so we need to convert to electrons here.
-            gain = exposure.getMetadata().get(f"LSST ISR GAIN {ampName}", np.nan)
+            gain = gains[ampName]
             overscanReadNoise = gain * metadataStats['READ_NOISE'][ampName]
             if overscanReadNoise:
                 if ((overscanReadNoise - readNoise)/readNoise > 0.05) or not np.isfinite(overscanReadNoise):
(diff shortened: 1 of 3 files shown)


python/lsst/cp/verify/verifyBias.py
```diff
@@ -150,11 +152,16 @@ def verify(self, exposure, statisticsDict):
             # consistently and significantly, then the assumptions
             # used in that test may be incorrect, and the nominal read
             # noise may need recalculation.  Only perform this check
-            # if the metadataStats contain the required entry.
-            overscanReadNoise = metadataStats['READ_NOISE'][ampName]
+            # if the metadataStats contain the required entry.  This
+            # is in ADU (the serial overscan is measured prior to gain
+            # normalization), so we need to convert to electrons here.
+            gain = gains[ampName]
+            overscanReadNoise = gain * metadataStats['READ_NOISE'][ampName]
```
erykoff: This is not correct.  The readnoise units are electrons https://github.com/lsst/ip_isr/blob/0d33ba980625e815f92b36fa222f8240fc568f41/python/lsst/ip/isr/isrTaskLSST.py#L1757C9-L1757C58 and you can look at the header to confirm this.  But now I'm disappointed that https://github.com/lsst/ip_isr/blob/0d33ba980625e815f92b36fa222f8240fc568f41/python/lsst/ip/isr/isrFunctions.py#L1616 doesn't make this clear.

erykoff (changes requested):
The units are incorrect which is maybe why ci_cpp is failing all tests?

python/lsst/cp/verify/verifyBias.py
erykoff: Okay, this code is correct but confusing because of the units.  Is it possible to ...
* rename the metadata stats (in dark and bias verification) to `READ_NOISE_ADU`
* Maybe call this `overscanReadNoiseElectron` (but if on the right it says `READ_NOISE_ADU` I think that would be obvious)
* Add a comment above on `readNoises` that this is returning the read noise in electrons.
* And in the dense `DMTN-101 Test 4.2: Mean is 0.0 within noise.` comment above say that "the nominal detector read noise is in electron units"

python/lsst/cp/verify/verifyBias.py
```diff
@@ -103,12 +104,13 @@ def verify(self, exposure, statisticsDict):
...
             A boolean indicating if all tests have passed.
         """
-        detector = exposure.getDetector()
         ampStats = statisticsDict['AMP']
         metadataStats = statisticsDict['METADATA']
 
         verifyStats = {}
         success = True
+        gains = getExposureGains(exposure)
+        readNoises = getExposureReadNoises(exposure)
```
erykoff: This might be clearer if this is `gainDict = ` and `readNoiseDict = `

python/lsst/cp/verify/verifyBias.py
```diff
@@ -129,7 +131,7 @@ def verify(self, exposure, statisticsDict):
             # measured on the bulk of the image as the overscan
```
erykoff: I can't comment into the block above but this should say that the nominal detector read noise is in electron, and if you update the field this should be `READ_NOISEA_ADU {ampName}`.

python/lsst/cp/verify/verifyDark.py
```diff
@@ -68,12 +69,13 @@ def verify(self, exposure, statisticsDict):
...
             A boolean indicating if all tests have passed.
         """
-        detector = exposure.getDetector()
         ampStats = statisticsDict['AMP']
         metadataStats = statisticsDict['METADATA']
 
         verifyStats = {}
         success = True
+        gains = getExposureGains(exposure)
+        readNoises = getExposureReadNoises(exposure)
```
erykoff: Same comments here across verifyDark as above.

erykoff (commented):
Now that I understand things better I think that we can rename some variable names to make it clearer when we look again in 6 months and have forgotten everything.

python/lsst/cp/verify/verifyBias.py
czwa: Agree on the first, left the second alone, as I think the changed key name makes that clear (and the preceding comment notes we're converting to electrons.  
Added comment above the gainDict/readnoiseDict noting that these are the PTC values used in constructing the variance plane (and that RN is in electrons).  
I've expanded the Test 4.3/5.3 comments a bit to note that this test runs against the PTC values.
I've clarified the Test 4.2/5.2 is comparing the mean to the noise measured on the image (which is in e-).
Hopefully clearer now (and perhaps in the future I will not name four different things "noise").

python/lsst/cp/verify/verifyDark.py
czwa: Edited in parallel so they should have identical changes, but I'll carefully check when I commit.

erykoff (approved):

PR #73, opened 2025-04-20
DM-50295: Update flat verification pipeline including CTI and excluding BF.

erykoff:


pipelines/_ingredients/verifyFlatLSST.yaml
```diff
@@ -16,12 +16,13 @@ tasks:
...
         config.doApplyGains = True
         config.doLinearize = True
-        # TODO DM-46426: Add cpCtiLSST pipeline so that this can be True.
-        config.doDeferredCharge = False
+        config.doDeferredCharge = True
         config.doBias = True
         config.doDark = True
         config.doDefect = True
-        config.doBrighterFatter = True
+        # TODO DM-50263: Investigate if this can be turned on, looking
```
czwa: I think this comment is wrong, because I went to make sure I was watching that ticket, but it's titled "Aperture Correction uncertainties": https://rubinobs.atlassian.net/browse/DM-50263

czwa (approved):
I think this has a ticket number typo.

PR #75, opened 2025-05-22
DM-50376: Add detailed plots

Alex-Broughton:


bin.src/cpv_report.py
```diff
@@ -22,7 +22,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 import sys
 
-from lsst.cp.verify.report import main
+from lsst.cp.verify.cpvReport import main
```
aferte, line 25: What is the motivation for renaming report? It seems redundant to me, but I also don't mind much - just wondering.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+from matplotlib.pyplot import cm
+from matplotlib.colors import SymLogNorm
+
+from lsst.daf.butler import Butler
+from lsst.daf.butler import EmptyQueryResultError, DatasetNotFoundError
+
+from .utils import plot_fp_statistic
+
+
+class DetailedPlotter():
```
aferte: are the parentheses needed here?

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+
+        # Get the camera associated with these calibrations
+        self.camera = self.butler.get(
+            "camera",
+            instrument=self.instrument,
+            collections=f"{self.instrument}/calib",
+        )
+
+        # Get the detector ids
+        self.detectorIds = range(len(self.camera)) if len(detectorIds) == 0 else sorted(detectorIds)
```
aferte, line 92: Shouldn't this also test for the existence of `detectorIds`? E.g. by adding `or detectorIds is None`.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+        return f"{raft}/{bay}"
+
+    def _getAllDictValues(self, d):
+        def _nested(_):
+            for v in _.values():
+                if isinstance(v, dict):
+                    yield from _nested(v)
+                else:
+                    yield v
+        return np.asarray(list(_nested(d)))
```
aferte, line 141: Why can't we just use `np.array(list(d.values()))` instead of this method?

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+                ax.set_title(ampName)
+                ax.scatter(ptc.photoCharges[ampName], ptc.rawMeans[ampName], s=1)
+                ax.axhline(lin.linearityTurnoff[ampName], linestyle="--", color="k", alpha=0.5)
+
+                ax.text(
+                    2e-7, 0.25e5,
+                    f"Linearity Turnoff:\n{round(linTurnoff, 1)} adu",
+                )
+
+                ax.set_xlim(-.5e-7, 6.5e-7)
```
aferte: A general comment for the rest of the code: is it ok to give specific values for the limits instead of limits dependent on the data? E.g. would these be good limits for LATISS?

python/lsst/cp/verify/templates/detailed.html
```diff
@@ -0,0 +1,218 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>name</title>
+    <style>
+        .grid-container {
+            display: grid;
+            grid-template-columns: repeat(15, 1fr);
```
aferte, line 8: This (and several other settings below) seems specific to LSSTCam. Is it possible to make it so we can use it for other cameras? I am thinking we will want to use this for LATISS as well in the future. If you prefer not to worry about other cameras, then I am fine but we should clarify in the code name and documentation that this is LSSTCam specific.

python/lsst/cp/verify/utils.py
```diff
@@ -18,27 +18,261 @@
...
+    "plot_detector",
+    "nsigma_range",
+    "get_median_nsigma_range",
+    "get_median_nsigma_range",
+    "plot_fp_statistic",
+]
+
+# All LSST cameras will have an R22. Therefore,
+# ComCam and LATISS will both map to the center
+# raft in the detailed plots.
```
aferte: OK maybe that addresses my previous comment. But does it turn out as a lot of blank space with a few plots in the middle for LATISS and ComCam?

aferte (commented):
Great work! I have a few comments in the review. 
Do you have examples I could look at? One for LSSTCam and one for ComCam or LATISS if possible? 
My main worry otherwise is that the general approach of cp verify is to use unified stats and plotting capabilities from analysis_tools but this is implementing its own plots and stats. If @czwa is fine with it, that's fine by me.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+    repo : `str`
+        The location of the butler repository to retrieve results
+        from.
+    instrument : `str`
+        The instrument associated with this data.
+    output_path : `str`
+        The location the report will be written to.
+    collections : `list` [`str`]
+        A list of collections to search.
+    **kwargs :
```
czwa, line 59: I don't think this is true.  `doOverwrite` will not be set if `kwargs['do_overwrite']` is found, and `do_copy` gets lost entirely.
`detectorIds` and `plotSummaryStats` are also not included here.

python/lsst/cp/verify/detailedPlots.py
czwa: This recurses into `d = {'my_favorite_stat' = 1.2345, 'all_stats': {'stat0': 0.0, 'stat1': 1.0, 'stat2': 2.0}}` to return `[1.2345, 0.0, 1.0, 2.0]`, although I suspect it's there to handle `d = {'detectorNameA': {'ampName0': 0.0, 'ampName1': 1.0}, 'detectorNameB': {'ampName0': 0.0, 'ampName1': -1.0}}`.

python/lsst/cp/verify/detailedPlots.py
czwa: `detectorIds` is a required argument, so it will exist.  This is possibly unsafe against `detectorIds="1,2,3,cat,dog"`, because strings have `len != 0`, but for a class that is unlikely to ever be randomly  called outside of the script, I don't think it's worth worrying about.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+        # Initialize output directories
+        for detector in self.camera:
+            detName = detector.getName()
+            raft = detName[0:3]
+            bay = detName[4:]
+            os.makedirs(f"{self.output}/{raft}", exist_ok=True)
+            os.makedirs(f"{self.output}/{raft}/{bay}", exist_ok=True)
+
+    def plotLinearizerStatistics(self):
+        self.plotIdx += 1
```
czwa, line 171: This leaves plots indexed starting at 1, not zero, if I'm thinking through the logic correctly, right?

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+                instrument=self.instrument,
+                where=f"detector in ({self.detectorIdsString})",
+                collections=self.collections,
+            )
+        except EmptyQueryResultError:
+            return
+
+        self.refs = refs
+
+        for ref in tqdm(self.refs, total=len(self.refs), desc="LINEARIZER"):
```
czwa: This needs a comment explaining what's happening.  It's reasonably simple to follow once I know what `tqdm` does, but I had to google it (although it is in the default stack install, so maybe it's just me).

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+                ax = axs[i]
+                ax.set_title(ampName)
+
+                if ampName in ptc.badAmps:
+                    ax.set_facecolor('#fcd4d4')
+                    ax.set_title(ampName + " (BAD)")
+
+                ptcTurnoff = ptc.ptcTurnoff[ampName]
+
+                data01 = ptc.covariances[ampName][:, 0, 1]
```
czwa: Typo?  `data01` and `data10` look like they're pulling the same elements out.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+            fig, axes = plt.subplots(4, 4, figsize=(12, 12))
+            fig.suptitle(self.camera[ref.dataId['detector']].getName() + f" ({ref.dataId['detector']})")
+            axs = axes.ravel()
+            detName = self.camera[ref.dataId['detector']].getName()
+            for i, amp in enumerate(self.camera[ref.dataId['detector']].getAmplifiers()):
+                ampName = amp.getName()
+
+                # Save dataset information
+                self.typesDict[detName][ampName] = self.camera[ref.dataId['detector']].getPhysicalType()
+                # Need to fix this. Why is it a zero dimensional array?
```
czwa: I filed DM-51582 to audit our calibrations to fix this.  I've been thinking about a larger `IsrCalib` refactor, but this would be a simple fix to make sure that the components of calibrations have reasonable types.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+            axs = axes.ravel()
+            for i, amp in enumerate(self.camera[ref.dataId['detector']].getAmplifiers()):
+                ampName = amp.getName()
+                ax = axs[i]
+                ax.set_title(ampName)
+                ax.scatter(ptc.photoCharges[ampName], lin.fitResiduals[ampName]/ptc.rawMeans[ampName], s=1)
+                ax.axhline(0, linestyle="--", color="k", alpha=0.5)
+                ax.ticklabel_format(style='sci', axis='x', scilimits=(0, 0), useMathText=True)
+                ax.set_ylim(-0.00075, 0.00075)
+                if ref.dataId['detector'] in [60, 61, 62]:
```
czwa: Like the above comment, I'm don't really like hard-coding limits, but I think these exceptions are going to be more of a problem.  It looks like this changes the limits from +/-x to +/-4x.  Can we add a sigma (or max-min difference) check to know if we should be in the +/-x or +/-4x range?  Then if some other detector goes bonkers, we won't get fooled because the plot is too narrow to see how bad the residuals are.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+
+            fig, axes = plt.subplots(4, 4, figsize=(12, 12))
+            fig.suptitle(self.camera[ref.dataId['detector']].getName() + f" ({ref.dataId['detector']})")
+            axs = axes.ravel()
+            for i, amp in enumerate(self.camera[ref.dataId['detector']].getAmplifiers()):
+                ampName = amp.getName()
+                ax = axs[i]
+                ax.set_title(ampName)
+
+                ptcTurnoff = ptc.ptcTurnoff[ampName]
```
czwa: I think this is unused?  Flake8 may not notice because it's used in prior `for` loops.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+
+        # Initialize output directories
+        for detector in self.camera:
+            detName = detector.getName()
+            raft = detName[0:3]
+            bay = detName[4:]
+            os.makedirs(f"{self.output}/{raft}", exist_ok=True)
+            os.makedirs(f"{self.output}/{raft}/{bay}", exist_ok=True)
+
+    def plotLinearizerStatistics(self):
```
czwa, line 170: I was fine with this one not having a docstring, but after reading the PTC method, I changed my mind.  I don't think we need a "This method plots linearizer statistics." line, but a `Notes` that lists what output plots are generated, with file templates, would be very useful if I don't remember the difference between `ptc_correlations_C00_` or `ptc_residuals_C00_`.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+                ax.set_ylabel("Parallel")
+            plt.tight_layout()
+            outputFigPath = f"{self.output}/{self._getRaftBayPathFromRef(ref)}"
+            plt.savefig(f"{outputFigPath}/step_{self.plotIdx:02}j-"
+                        f"ptc_noise_matrix_{self.camera[ref.dataId['detector']].getId():03}.png",
+                        bbox_inches='tight')
+            plt.close()
+
+    def plotBfkStatistics(self):
+        self.plotIdx += 1
```
czwa, line 710: This increments the index, even if no files are found.  Is this to ensure that "step" labels in the filenames are the same, regardless of the data?

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+                ax = axs[i]
+                ax.set_title(ampName)
+                ax.scatter(cti.signals[ampName], cti.serialEper[ampName], s=1)
+
+                globalCti = cti.globalCti[ampName]
+                serialCtiTurnoff = cti.serialCtiTurnoff[ampName]
+
+                ax.axhline(0, linestyle="--", color="k", alpha=0.5)
+
+                try:
```
czwa, line 803: Should `num` be `globalCti / (10**np.abs(power))`?  Running some tests indicates this doubles the order:

```
gc = [-3.2e6, 3.2e-6]
pp = np.floor(np.log10(np.abs(gc)))
nn = gc * 10**(pp)
print(gc, pp, nn)
[-3200000.0, 3.2e-06] [ 6. -6.] [-3.2e+12  3.2e-12]
```

python/lsst/cp/verify/detailedPlots.py
czwa: I had a comment here concerned about this extending into the non-science detectors, but I think this is resolved by the try/excepts on the ref getting below.

python/lsst/cp/verify/detailedPlots.py
```diff
@@ -0,0 +1,1266 @@
...
+        # Plot
+        with logging_redirect_tqdm():
+            self.plotLinearizerStatistics()
+            self.plotPtcStatistics()
+            self.plotBfkStatistics()
+            self.plotCtiStatistics()
+            self.plotBiasStatistics()
+            self.plotDarkStatistics()
+            self.plotBiasVsDarkStatistics()
+            if self.plotSummaryStatistics:
```
czwa: I don't think I understand this.  If this method exists, run it?  When would it not exist?

python/lsst/cp/verify/templates/detailed.html
```diff
@@ -0,0 +1,218 @@
...
+</head>
+<body>
+    <h1>name</h1>
+    <div class="image-container">
+        <div class="grid-container">
+            grid_placeholder
+        </div>
+    </div>
+</body>
+</html>
```
czwa, line 218: This at least needs the missing newline fixed (and in `summary.html`).

python/lsst/cp/verify/utils.py
czwa: LATISS doesn't have an R22.  LATISS's single detector is `RXX_S00`.

python/lsst/cp/verify/utils.py
```diff
@@ -50,3 +284,281 @@ def mergeStatDict(statDictA, statDictB):
...
+    amps: container-type object [None]
+        Python container that can be queried like `'C01 in amps'`
+        to see if a particular channel is included for plotting.
+        If None, then use all channels in det.
+
+    Returns
+    -------
+    list of matplotlib.patches.Rectangle objects
+    """
+    transform = det.getTransform(cameraGeom.PIXELS, cameraGeom.FOCAL_PLANE)
```
czwa, line 308: This doesn't correctly handle rotations.  See: https://github.com/lsst/analysis_tools/blob/main/python/lsst/analysis/tools/actions/plot/focalPlanePlot.py#L465

czwa: In fact, I think I'm going to suggest that if these functions are essential, they should be factored into methods on `FocalPlaneGeometryPlot`, so that there's only one code path doing this.

python/lsst/cp/verify/detailedReportsLSSTCam.py
```diff
@@ -0,0 +1,265 @@
...
+    repo : `str`
+        The location of the butler repository to retrieve results
+        from.
+    instrument : `str`
+        The instrument associated with this data.
+    output_path : `str`
+        The location the report will be written to.
+    collections : `list` [`str`]
+        A list of collections to search.
+    **kwargs :
```
czwa, line 54: Same comment as above on this docstring not matching the code.

python/lsst/cp/verify/detailedReportsLSSTCam.py
```diff
@@ -0,0 +1,265 @@
...
+        self.instrument = instrument
+        self.doOverwrite = doOverwrite
+
+        # Check that the output directory
+        # is organized in the proper format
+        if not os.path.isdir(self.sharedPath):
+            raise RuntimeError(f"The shared directory {self.sharedPath} does not exist.")
+        if not os.path.isdir(self.sharedPath + "/images"):
+            raise RuntimeError(f"The shared directory {self.sharedPath} does not follow the expected "
+                               "organizational structure, please refer to DMTN-222 for more information.")
```
czwa, line 73: Is there DMTN-222 ticket branch with this info?

python/lsst/cp/verify/detailedReportsLSSTCam.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def getDetailedImages(self):
+        tree = defaultdict(list)
+        # Get tree data
+        fullNames = []
+        for root, dirs, files in os.walk(self.sharedPath):
+            for file in files:
+                if file.endswith('.png') and not file.endswith('_fp.png'):
+                    try:
+                        pattern = re.compile(r"(step_\d{2}[a-zA-Z]?)-([a-zA-Z0-9_-]+)_(\d{2,3})\.png")
```
czwa, line 132: I have concerns.

czwa (commented):
Documentation requests, some clarifications where I got confused, and at least one likely typo (the C01/C10 covariance terms).

The concerns mentioned in `detailedReport.py`:
1. How long does this take to run to generate plots?  The previous version of the `cpv_report` also attempted to generate all of the plots at once, and it took an infeasible amount of time to run.  This led to the `analysis_tools`-ification, as then we could have tasks to generate all the plots that run as part of the pipeline.
2. This also solves the issue with `getDetailedImages`.  If the plots are generated as part of the pipeline, then they have butler dataIds themselves, and so each plot type (corresponding to a dataset_type in the butler) can be directly requested from the `refs`, and used to build the report.
3. I'm generally worried about diverging cpv reports.  I'm happy to switch the existing plots into this report format, as the existing report is very ugly and difficult to navigate.  However, I think the base idea of "generate in task, collate by script" is more robust because it keeps most of the work in something that's more parallelizable.

python/lsst/cp/verify/utils.py
```diff
@@ -18,27 +18,261 @@
...
+}
 
 
 def mergeStatDict(statDictA, statDictB):
     """Merge one dictionary of statistics with another.
 
     Parameters
     ----------
-    statDictA, statDictB : `dict` [`str`, `dict`]
+    statDictA, statDictB: `dict` [`str`, `dict`]
```
czwa, line 263: The dev guide claims these are important spaces: https://developer.lsst.io/python/numpydoc.html#parameters

python/lsst/cp/verify/utils.py
```diff
@@ -18,27 +18,261 @@
...
-    output : `dict` [`str`, `dict`]
+    output: `dict` [`str`, `dict`]
         A dictionary containing the union of the internal
         dictionaries of ``statDictA`` and ``statDictB``, indexed
         by the amplifier names.
 
     Raises
     ------
-    RuntimeError :
+    RuntimeError:
```
czwa, line 275: But the dev guide also suggests this shouldn't have a `:` at all, so I'm learning new things too: https://developer.lsst.io/python/numpydoc.html#parameters

bin.src/cpv_report.py
Alex-Broughton: I renamed it so that there is less confusion between the two different "reports"

python/lsst/cp/verify/detailedPlots.py
Alex-Broughton: For the detailed plots, we want to make sure everything has the same limits, so that when viewing everything together, one can easily find outliers.

python/lsst/cp/verify/templates/detailed.html
Alex-Broughton: The detailed viewer is meant for finding outliers in statistics for large numbers of sensors, like in LSSTCam. I don't think we will need the detailed web pages for LATISS. That is why I separated out the plot generation from the web reports. We can make all plots for every instrument, but we only need to make the web reports for LSSTCam. This would work for AuxTel and ComCam though.

python/lsst/cp/verify/utils.py
Alex-Broughton: Ah, ok. I might have to fix this then

python/lsst/cp/verify/templates/detailed.html
Alex-Broughton: Ok, I updated it so that the web reports are only for LSSTCam. ComCam's done and we won't need the web reports for AuxTel since it is just one sensor.

python/lsst/cp/verify/detailedPlots.py
Alex-Broughton: That's right. I wanted to leave it indexed at zero as it is more human-readable

PR #80, opened 2025-11-25
DM-52350: Make sure that linearityType is always a python unicode string

Alex-Broughton:


czwa (approved):
This was a simple one.

PR #81, opened 2026-01-14
DM-51599: Run crosstalk on CBP data / update and modernize CT code 

czwa:


erykoff (approved):
I think this is fine, but have you actually run this verify code?

czwa:
I don't think I've run the whole pipeline since I started the crosstalk work.  I'm going to run a confirmation check before merging this ticket.

PR #82, opened 2026-03-06
DM-53191: add `MeasureDefectsStability`

seanmacb:
This is a change to add a new task to `verifyDefects.py`, providing a tool to measure variation in defect masks. 

This addresses [DM-53191](https://rubinobs.atlassian.net/browse/DM-53191)

[DM-53191]: https://rubinobs.atlassian.net/browse/DM-53191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pipelines/_ingredients/verifyDefectStabilityLSST.yaml
```diff
@@ -0,0 +1,24 @@
+description: cp_verify for Defect stability across combined calibrations.
+imports:
+  - location: $ANALYSIS_TOOLS_DIR/pipelines/cpCore.yaml
+    include:
+      - analyzeDefectCore
```
czwa, line 5: Does this actually use `analyzeDefectCore`?  I had to look it up, but it appears to be making focal plane plots of N_{defect pixels} and fraction_{defect pixels}.  If the output table doesn't have the correct columns, that task will fail.
I'll return to this once I've read through the tasks.

python/lsst/cp/verify/verifyDefects.py
```diff
@@ -341,12 +347,18 @@ def repackStats(self, statisticsDict, dimensions):
...
+                    )
                 elif ampName in stats.keys():
                     key = f"LSST CALIB DEFECTS {ampName} N_HOT"
-                    rows[ampName][f"{self.config.stageName}_N_HOT"] = stats[ampName].get(key, np.nan)
+                    rows[ampName][f"{self.config.stageName}_N_HOT"] = stats[
+                        ampName
+                    ].get(key, np.nan)
                     key = f"LSST CALIB DEFECTS {ampName} N_COLD"
-                    rows[ampName][f"{self.config.stageName}_N_COLD"] = stats[ampName].get(key, np.nan)
+                    rows[ampName][f"{self.config.stageName}_N_COLD"] = stats[
```
czwa, line 359: Are these automatic formatting fixes (from black or something similar)?

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+)
+from .verifyStats import (
+    CpVerifyStatsConfig,
+)
+
+__all__ = ["CpMeasureDefectsStabilityTaskConfig", "CpMeasureDefectsStabilityTask"]
+
+
+def lookupStaticCalibrations(datasetType, registry, quantumDataId, collections):
+    # This will retrieve all the filter transmissions available
```
czwa, line 41: Typo/miscopy/etc.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+):
+    """
+    Configuration for measuring stability
+    of defects from individual defect masks.
+    """
+
+    refLabel = pexConfig.Field(
+        dtype=str,
+        doc="Label to apply to the reference defect mask.",
+        optional=False,
```
czwa, line 114: This and the next should have defaults if they're not optional.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+
+    skipMasks = pexConfig.ListField(
+        dtype=str,
+        doc="Mask planes to skip.",
+        default=(),
+        optional=True,
+    )
+
+    includeAny = pexConfig.ChoiceField(
+        doc="Flag to include the any masks in variability computation.",
```
czwa, line 131: I made a stub comment here to return to after reading the code, but I'm confused as to how `processDetectorsAndAmps` works, so I don't think I have anything to add here other than "I'm confused as to how this works."

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+            True: "Any mask variability is included in the computation.",
+            False: "Any mask variability is excluded in the computation.",
+        },
+        default=True,
+        optional=True,
+    )
+
+    def setDefaults(self):
+        # Set the acq string based on the two labels
+        self.acqString = ""
```
czwa, line 143: Why is this happening in the config object?  This seems like it should be part of the task itself.  I'm not even sure you have access to the camera here.  I would have thought this would execute when the config is created, and the camera should be readable until the task's `runQuantum` method (where the references are found).

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+
+    ConfigClass = CpMeasureDefectsStabilityTaskConfig
+    _DefaultName = "cpMeasureDefectsStability"
+
+    def evaluate_mask_variability(self, masks, dicts):
+        """
+        Evaluate the variability of each mask plane across an entire
+        detector/amplifier
+
+        masks: List of 2D numpy arrays, each corresponding to the
```
czwa, line 188: These should follow the standard docstring format (https://developer.lsst.io/python/numpydoc.html):
```
Parameters
----------
masks : `list` [`np.ndarray`, (nRows, nCols)]
    Words.
dicts : `list` [`dict` [`str`, `int`]]
    Words.

Returns
-------
stacks : `dict` [`str`, `np.ndarray`, (nMasks, nRows, nCols)]
    Words.
```

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+                        defectData.loc[f"{detectorName}_{ampName}", f"{acq}_Any"] = (
+                            np.count_nonzero(anyMasks[i])
+                        )
+                    defectData.loc[
+                        f"{detectorName}_{ampName}", f"{self.config.acqString}Any"
+                    ] = np.sum(anyMasks[i - 1] - anyMasks[i], dtype=int)
+
+        return defectData
+
+    def runQuantum(self, butlerQC, refRefs, prodRefs, outputRefs):
```
czwa, line 321: You don't need a `runQuantum` if you're not doing any operations there (the inherited default does this for you).

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+        prodInput = butlerQC.get(prodRefs)
+        refInput = butlerQC.get(refRefs)
+
+        outputs = self.run(prodInput, refInput)
+        butlerQC.put(outputs, outputRefs)
+
+    def run(self, referenceDefects, productionDefects, defectData):
```
czwa, line 328: Docstring.  Also, looking at how this is called in `runQuantum`, you've flipped the production and reference defects, and `defectData` isn't defined.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+        return pipeBase.Struct(
+            self.processDetectorsAndAmps(
```
czwa, line 330: This needs a keyword that matches the output connection:
```
return pipeBase.Struct(
    outputTable=self.processDetectorsAndAmps() #args skipped for this comment
)
```

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+
+                planes.append(binary_plane.astype(float))
+
+            # Convert list to a 3D stack (N, Height, Width)
+            stack = np.stack(planes)
+
+            stacks[name] = stack
+        return stacks
+
+    def processDetectorsAndAmps(self, referenceDefects, productionDefects, defectData):
```
czwa, line 223: Docstring.  I'd also suggest just putting all of this into the `run` method.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+        for detectorName, detectorObject in self.config.detName_Obj.items():
```
czwa, line 224: This needs a unit test, because as mentioned in the `config.setDefaults`, I don't know that this will work.  The default `runQuantum` will ensure that there's a `camera` object supplied to `run`, so all of the work done in `setDefaults` can be moved to `run`.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+            if all((mask1.array - mask2.array).flatten() == 0):
+                print("Mask planes are identical.")
+            else:
+                print("Mask planes are not identical.")
+            # NOTE: Commenting the above out for now, there is definitely
+            # a faster way to address identical mask planes...
+            """
+
+            # Assert that the masks have the same plane definitions
+            assert (
```
czwa, line 263: The `maskPlaneDict` is a singleton (much to my sadness), so by definition they should be identical.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+               detector/amplifier.
+        dicts: List of bit_position dictionaries corresponding to
+               each mask.
+        """
+        # Find the union of all defect names across all masks
+        all_defect_names = set().union(*(d.keys() for d in dicts))
+
+        stacks = {}
+
+        for name in all_defect_names:
```
czwa, line 199: This will currently only ever populate the `BAD` plane, as that's all that's stored in the `Defects`.  There are reason planes coming to identify what kind of `BAD` the pixel is, but that hasn't merged.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+                mask1.getMaskPlaneDict() == mask2.getMaskPlaneDict()
+            ), "Mask plane dictionaries are incompatible"
+
+            # For each amp
+            for amp in detectorObject.getAmplifiers():
+
+                if self.config.includeAny:
+                    # Instantiate the anymask object
+                    anyMasks = []
+                    for acq in [self.config.refLabel, self.config.prodLabel]:
```
czwa, line 273: I don't understand this.  This makes two full-zero arrays the shape of a given amp, but then doesn't do anything with them.  It's used later on L296, but if `includeAny` is false, that's going to result in a `NameError`.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+                )
+
+                # For each mask plane
+                for plane, planeVals in variability.items():
+                    if plane not in self.config.skipMasks:
+                        # Compute NPix for each run and add it to the df
+                        for i, acq in enumerate(
+                            [self.config.refLabel, self.config.prodLabel]
+                        ):
+                            defectData.loc[
```
czwa, line 291: I don't see where this is ever declared.  `run` assumes it's an input, but it doesn't exist in `runQuantum`.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+):
+    """
+    Connections for plotting stability
+    of defects from individual defect masks.
+    """
+
+    inputTable = cT.PrerequisiteInput(
+        name="defectsStability",
+        doc="Table of defect stability measurements.",
+        storageClass="DataFrame",
```
czwa, line 347: Can I convince you to make this an `ArrowAstropy` instead?  I find astropy tables much easier to deal with than Pandas.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+    )
+
+    camera = cT.PrerequisiteInput(
+        name="camera",
+        doc="Camera associated with these defects.",
+        storageClass="Camera",
+        dimensions=("instrument",),
+    )
+
+    outputTable = cT.Output(
```
czwa, line 358: If this is a plot, it shouldn't be called a table.  I also think this maybe should be in `analysis_tools`, as then it's clear that we want to use that apparatus to get these plots into the plot navigator.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+    """
+
+    skipPlanes = pexConfig.ListField(
+        dtype=str,
+        doc="Mask planes to skip when plotting.",
+        default=(),
+        optional=True,
+    )
+
+    _filter = pexConfig.Field(
```
czwa, line 382: I don't know that we have private config options, so I think this is better without an underscore.  Unless it's a typo.  In which case it's still better without the underscore.

python/lsst/cp/verify/verifyDefectStability.py
```diff
@@ -0,0 +1,505 @@
...
+        dtype=str,
+        doc="Filter that the defects were generated from, used in plotting.",
+        default="None",
+        optional=True,
+    )
+
+    tract = pexConfig.Field(
```
czwa, line 389: One day I'll work on making these not required.  It's annoying, but everyone else who does `analysis_tools` work works on downstream products that have these dimensions.

czwa (changes requested):
Sorry for the many comments, but there are too many parts that I just don't understand.  Main concerns:

1. Docstrings.  I know their tedious to write, but they might have helped me understand what's going on.
2. Someone will complain if there are camel and snack case functions in the same task.
3. The `config.setDefaults` is doing things I don't think should work.
4. I think the plotting should be in `analysis_tools`.
5. Unit tests would be good, or, barring the difficulty of setting up everything, `ci_cpp_gen3` should exercise this code (I think there should be two sets of defects after the defects step runs: the original curated ones, and the ones generated from the dark/flat that are a superset of the curated ones).

PR #85, opened 2026-06-15
DM-54582: Add tooling to generate the detailed + summary interactive plotting webpage

Alex-Broughton:


python/lsst/cp/verify/interactiveReportLSSTCam.py
```diff
@@ -0,0 +1,261 @@
...
+    collections : `list` [`str`]
+        A list of collections to search.
+    **kwargs :
+        Other keyword parameters.  Currently parsed values:
+
+        - ``do_copy``: Should files be copied from butler (`bool`).
+        - ``do_overwrite``: Should pre-existing files be overwritten (`bool`).
+    """
+
+    def __init__(self, sharedPath, doOverwrite, **kwargs):
```
czwa: The parameters in the `Parameters` docstring are not the parameters here.  Did the docstring get out of sync with the code at some point?

python/lsst/cp/verify/interactiveReportLSSTCam.py
```diff
@@ -0,0 +1,261 @@
...
+        super().__init__()
+        # Set source and destination information.
+        self.sharedPath = sharedPath
+        self.doOverwrite = doOverwrite
+
+        # Check that the output directory
+        # is organized in the proper format
+        if not os.path.isdir(self.sharedPath):
+            raise RuntimeError(f"The shared directory {self.sharedPath} does not exist.")
+        if not os.path.isdir(self.sharedPath + "/images"):
```
czwa, line 63: Is there a DMTN-222 PR?  I couldn't find the information on what the expected pattern is supposed to be.  It looks like this expects `/path/to/my/directory` and `/path/to/my/directory/images` to both exist.  Is that correct?

python/lsst/cp/verify/interactiveReportLSSTCam.py
```diff
@@ -0,0 +1,261 @@
...
+                    histImages.append(f)
+                elif f.endswith('_mosaic.png') and os.path.isfile(os.path.join(imagePath, f)):
+                    mosaicImages.append(f)
+                else:
+                    continue
+
+        mosaicImages = sorted(mosaicImages)
+        histImages = sorted(histImages)
+
+        summaryImages = list(zip(mosaicImages, histImages))
```
czwa, line 118: This is likely me being overly cautious, but can this check that the lengths are the same before this line?

python/lsst/cp/verify/interactiveReportLSSTCam.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Instantiate a butler for our repository.
+        self.butler = Butler("/repo/main")
+        self.registry = self.butler.registry
+
+        # Get the camera associated with these calibrations
+        self.camera = LsstCam.getCamera()
+
+    def writeReport(self, renderedHtml, fileName):
+        output_path = self.sharedPath + "/" + fileName
```
czwa, line 80: Can these methods get docstrings?  It's easier for me (at least) to understand how everything connects with docstrings.

python/lsst/cp/verify/interactiveReportLSSTCam.py
```diff
@@ -0,0 +1,261 @@
...
+                    <img src="{1}" alt="Hist Image {2}">
+                </a>
+            """.format(mosaicPath, histPath, i)
+
+        renderedHtml = summaryReportTemplate.replace("summary_images", addition)
+
+        self.writeReport(renderedHtml, "summary.html")
+
+    def makeDetailed(self, fullName, tree):
+        # Get the template for the detailed report
```
czwa, line 201: Above docstring comment prompted by this.  This expects `fullName` and `tree`, but this `tree` is not the `tree` from `getDetailedImages`, right?  It's the `grid` list-of-lists, right?

czwa (changes requested):
Almost entirely docstring requests.  I think I follow everything in python, and just accept that the html does what it needs to.

python/lsst/cp/verify/interactiveReportLSSTCam.py
```diff
@@ -0,0 +1,261 @@
...
+    parser.add_argument("--do-overwrite", dest="doOverwrite", action="store_true", default=False,
+                        help="Allow existing files to be overwritten?")
+    args = parser.parse_args()
+
+    reporter = DetailedReporterLSSTCam(
+        sharedPath=args.sharedPath,
+        instrument=args.instrument,
+        doOverwrite=args.doOverwrite,
+    )
+    reporter.run()
```
czwa, line 305: ETA: I think I actually want "Approved, pending docstrings are updated", so this is my re-review to change the review state.

czwa (approved):
Approved, pending docstrings are added/updated.

PR #86, opened 2026-09-04
DM-51000: Prepare for eups less installations

mwittgen:
DM-51000: Prepare for eups less installations


python/lsst/cp/verify/resources/pipelines
```diff
@@ -0,0 +1 @@
+../../../../../pipelines
```
czwa: I was going to ask about the missing newline, but since this is a github symlink (something I don't think I've seen before), I don't know that there's a fix.

timj: @mwittgen we will have to update the developer guide once we have made the changes

